PR #6, opened 2020-07-04
Add register aliases : `RegisterEntryPoint`, `RegisterComponent`

hadashiA:
Add new two register methods

## `RegisterEntryPoint<T>`

To register entry point markers, (e.g: `IInitializable`, `ITickable`, etc) we used this until now:

```csharp
builder.Register<Foo>(..).As<ITickable>();
// Or
builder.Register<Foo>(...).AsImplementedInterfaces();
```

- I think this is explicit, but a bit unclear about side effects
- This is a bit frequently, so I think it is better to have a shorter hand.

I will add following:

```csharp
builder.RegisterEntryPoint<Foo>(..);
```

## `RegisterComponent<T>`

In the following cases, I noticed that Inject doesn't work even though I have an instance of MonoBehaviour.

```csharp
public MyMonoBehaviour : MonoBehaviour
{
    [Inject]
    public void Construct(ServiceA a) { /* ... */ }
}

public class MyInstaller : MonoInstaller
{
    [SerializeField]
    MyMonoBehaviour component;

    public void Install(IContainerBuilder builder)
    {
        builder.RegisterInstance(component);

        // Nothing to other.
    }
}
```

In this case, MyMonoBehaviour registered, but nothing to resolve, then nothing inject.
To fix this, Inject now works when Registering MonoBehaviour.

However, container needs to know that the registered obj is MonoBehaviour. Therefore, I added RegisterComponent instead of RegisterInstance.

```csharp
    [SerializeField]
    MyMonoBehaviour component;

    // ...

    builder.RegisterComponent(component);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -582,8 +582,6 @@ class OtherClass
 
 **Register from MonoInstaller's `[SerializeField]`**
 
-Also use `RegisterInstance()`.
-
 ```csharp
 [SerializeField]
 YourBehaviour yourBehaviour;
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 
 ![](docs/unity_performance_test_result.png)
 
-- This benchmark was run by [Performance Testing Extension for Unity Test Runner](https://docs.unity3d.com/Packages/com.unity.test-framework.performance@1.0/manual/index.html).  Test cases is [here](./VContainer.Benchmark).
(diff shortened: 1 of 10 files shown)


hadashiA:
Refs: [Unityの新しいDIライブラリ：VContainer (7/3修正) - Qiita](https://qiita.com/__poosuke/items/bb8b91202ca5ba6229dc)

PR #35, opened 2020-10-09
Fix a problem with duplicate instances when resolving a Singleton  from a child scope

hadashiA:
Singleton in the root container had to be taken out of the Root container.


Watcher3056:
Hi!

How to avoid creating singleton in the child scope when parent(or Root) scope already has instance of this singleton?

PR #121, opened 2021-02-11
Restore prefab state after instantiation

cooloon:
Instantiated prefab itself (not clone in the scene) remains inactive even after play mode.
This PR restore prefab state just after instantiation.

If that's not inteded, just drop this PR 😃


hadashiA:
Oh.. ok. You are right, it's not good that the prefab remains inactive.

I would like to add a little work to it.
- Add some tests
- I'd like to reduce the number of calls to SetActive a bit (since the callback in the hierarchy is called) 

However, I would like to merge this. Thanks!

Issue #11, opened 2020-07-05
support for dynamic objects / factories and generic unconstructed types


slimshader:
Hey, VContainer looks great but looking at the docs it seems it doesn’t currently support spawning of objects at runtime via Factories, is that the case? Same question for unconstructed generic types: for example is there a way to bind all types that implement interface IEventHandler<T> from specified Assembly?

hadashiA:
@slimshader 

> it seems it doesn’t currently support spawning of objects at runtime via Factories, is that the case?

Hmm, I know many libraries support this. It deserves consideration.

But currently not supported.

I think it is not very important thing to inject to dynamic spawned object at runtime.

There are two reasons:
- Perhaps the objects created by the Factory are not of service class or any domain functionality, they're more granular. Therefore, I think a bit out of scope from a pure DI pattern.
- It may be possible to implement manually.

```csharp
class FooFactory
{
    readonly Settings settings;

    [Inject]
    public FooFactory(Settings settings)
    {
        this.settings = settings;
    }
 
    public Foo Create()
    {
        var foo = // ...
        foo.Configure(settings);
        return foo;
    }
}
```

Well, this is a bit redundant.
I might implement in the future if I come up with good API.

> is there a way to bind all types that implement interface IEventHandler from specified Assembly?

It also does not supported.
Is the function you are referring to something like the following ?

```        
    foreach (Type type in AllTypesInAssembly) 
    { 
        if (type.GetGenericTypeDefinition() == typeof(IEventHandler))
        {
            builder.Register(type)......
         }
    }
```

This is easy to implement, but performance seems bad.
Let me think about whether to implement it.

If you can tell me your use case, I may raise the priority.

slimshader:
> @slimshader
> 
> > it seems it doesn’t currently support spawning of objects at runtime via Factories, is that the case?
> 
> Hmm, I know many libraries support this. It deserves consideration.
> 
> But currently not supported.
> 
> I think it is not very important thing to inject to dynamic spawned object at runtime.
> 
> There are two reasons:
> 
> * Perhaps the objects created by the Factory are not of service class or any domain functionality, they're more granular. Therefore, I think a bit out of scope from a pure DI pattern.
> * It may be possible to implement manually.
> 
> ```cs
> class FooFactory
> {
>     readonly Settings settings;
> 
>     [Inject]
>     public FooFactory(Settings settings)
>     {
>         this.settings = settings;
>     }
>  
>     public Foo Create()
>     {
[...]

PR #153, opened 2021-03-01
Expose CreateInstance for Container

piti6:
What I want to do in RegisterFactory closure:
- Resolve constructor parameters automatically from container registry

What I can do without current pull request:
- Resolve constructor parameters manually by using `Container.Resolve<T>`, one by one.
- Create instance first, then call `Container.Inject`, which can't be applied to constructor parameters

What I can do with current pull request:
- Resolve constructor parameters automatically by using `Container.CreateInstance<T>`

Why do I need to expose InjectParameter.cs?
- cause `Match(Type parameterType, string parameterName)` has only type or name parameters, which means there's no need to define another derived class from outside as method has limited usage.

Why do I remove `Inject(object instance)`?
- cause you can call `Inject(object instance, params IInjectParameter[] parameters);` without parameters - like `Inject(instance)`, and it changes into `Inject(instance, System.Array.Empty<object>());` internally.


hadashiA:
Hi @piti6 . Thanks for the frequent PR!

I have two questions.

The first. `container.Resolve<T>()` automatically resolves the parameters of the T constructor.

The reason I often use the following example, is to make Resolve stable by making it at Factory creation time, not at T creation time.

```csharp
builder.RegisterFactory<Foo>(container =>
{
    RegisterFactory<Foo>(container => { var param1 = container.Resovle<Param1>(); // Resolve
    return () => new Foo(param1); // No Resolve
});
````

But the following is also possible

```csharp
builder.RegisterFactory<Foo>(container =>
{
    return () => contaienr.Resolve<Foo>();
})
```

Is this not enough for you ?

--

Second question. 

Why do you do dependency resolution at instance create time when you are using a Factory?

In my opinion, one of the purposes of the Factory pattern is to stabilize the timing of dependency resolution by moving it to "at Factory creation time instead of instance creation time".

Generally speaking, I believe that objects with dynamic lifetimes should not have complex "dependencies".

piti6:
> The first. container.Resolve<T>() automatically resolves the parameters of the T constructor.
Is this not enough for you ?

Oh, I did not noticed that I can resolve type without any register, (Auto wiring supported?)
but it does not helps my situation, as `container.Resolve` may only works when constructor does not have dynamic parameters.
I'm using factory pattern because I have dynamic parameter for constructor.

> Second question.
Why do you do dependency resolution at instance create time when you are using a Factory?
In my opinion, one of the purposes of the Factory pattern is to stabilize the timing of dependency resolution by moving it to "at Factory creation time instead of instance creation time".

Actually it doesn't matter to me that whether dependency resolution occurs at instance create time or factory creation time, (and I prefer design that dependency resolution timing does not matter, which means that injected instances has always same state and has no dependency to resolution timing.) as I'm using factory pattern to resolve dependencies with dynamic parameters in it - if CreateInstance is exposed, I may not need to use RegisterFactory also.

> Generally speaking, I believe that objects with dynamic lifetimes should not have complex "dependencies".

It might be, but I think registered instances (whether created by factory pattern or not, which created by lifetimeScope, such as injected singleton objects, etc, because they have scopes, and they are mutable since disposing scope would dispose instances.) are dynamic objects overall, not static.

piti6:
I realized that this PR might be resolve #11, as dynamic creation with resolving dependencies would be supported by using CreateInstance.

hadashiA:
>  but it does not helps my situation, as container.Resolve may only works when constructor does not have dynamic parameters.

Ah.. Ok.  It seems that you are right.
Auto-wiring is supported in Resolve<>, but you can't pass arguments dynamically.
And it doesn't work without Register.

>  I think registered instances (whether created by factory pattern or not, which created by lifetimeScope, such as injected singleton objects, etc, because they have scopes, and they are mutable since disposing scope would dispose instances.) are dynamic objects overall, not static.

Hmmm. If Resolve happens anytime, anywhere, then I don't think it's much different from Service Locator.
(I'm a bit of proponent of [Service Locator is an Anti-Pattern](https://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern/])

I know that Zenject brings the "Sub Container" and Factory patterns together, but this is very distinctive.
In a web application, the "request scope" is generated at the time of request. This is not a Factory pattern.

The DI pattern is not "you never know when Resolve will happen at runtime", it is "Resolve is done at scope construction time".
This is what I mean by stable.
If you need a dynamic Inject, I think it is preferable to create the scope dynamically.

hadashiA:
I thought about your suggestion again.
Presumably, this PR will allow you to do the following?

```csharp
container.CreateInstance<Foo>(
    new TypedParameter(a), 
    new TypedParameter(b));
```

I don't think that this is inherently more useful than what is currently possible.

```csharp
new Foo(a, b, container.Resolve<C>());
```

The difference between them is whether Resolve is implicit or not.
- But as long as you don't need a large number of arguments, I don't see much difference.
- And if you need a lot of arguments, I guess you could split the instance.

Also, `new TypedParameter(...) ` doesn't seem to be very concise.

For these reasons, I would like to reject this.
But thanks for your suggestion.

piti6:
@hadashiA 
> The difference between them is whether Resolve is implicit or not. 
- Difference becomes when you add more dependencies on already defined type, as explicit resolve needs to write **additional resolve** on installers, or validation will be fail.
- Also, **using constructor on resolve makes you to enforce to use constructor injection**, as field injection does not work in your way.

therefore, still there`s no way to resolve instance with field injection.

piti6:
> therefore, still there`s no way to resolve instance with field injection.

Oh, container.Inject might be OK I guess?

PR #156, opened 2021-03-02
Fix WithParameter does not work with property or field injection

piti6:
Subject:
```csharp
    public class Hoge
    {
        [Inject]
        private readonly int _hoge;
    }
    
    protected override void Configure(IContainerBuilder builder)
    {
        builder.Register<Hoge>(Lifetime.Singleton)
            .WithParameter(1234);
    }
```
Expected behaviour: 
- Builder should not throw any Exception

Actual result:
- VContainerException: No such registration of type: System.Int32


hadashiA:
When changing the behavior of the ReferectionInjector, I found that I needed to change the CodeGen mode as well.
https://github.com/hadashiA/VContainer/blob/master/VContainer/Assets/VContainer/Editor/CodeGen/InjectionILGenerator.cs#L326-L362

Let me think for a moment.

piti6:
Transfer to #457

Issue #167, opened 2021-03-27
TestRunner raises StackOverflowException when scripts are in Assembly-CSharp


shirokurohitsuji:
## Project
```
Assets/
└ Scenes/
    └ SampleScene.unity
└ Foo.cs
```

```
public class Foo { }
```

### Versions
- Unity Editor 2020.3.0f1
- VContainer v1.6.0
- Test Runner 1.1.24

## Description
**See below for the English version.**

1. MonoBehaviourを継承していないクラスをasmdefを切らずに作成する（Assembly-CSharpに含める）
2. VContainerをインストールする
3. UnityEditorの Window>General>Test RunnerからTest Runnerタブを開く

この一連の流れによって、StackOverflowExceptionが起きてから無限にNullReferenceExceptionが表示されるようになります。
コードをAssembly-CSharp以外のasmdefに収めることで解消します。
エラーが出ている間はTest Runnerがテストケースを読んでくれない状態になるため、Test Runnerが使えないという状態です。

バグなのかどうか、またバグだったとしてもVContainerのものなのか、TestRunnerのものなのかが掴めておらず申し訳ないのですが、確認していただければ幸いです。

1. create a class that does not inherit MonoBehaviour without compiling them into user's asmdef (include it into predefined Assembly-CSharp)
2. install VContainer
3. open the Test Runner tab from Window > General > Test Runner in Unity Editor

This raises a StackOverflowException and an infinite NullReferenceExceptions.
If you compile the codes into another asmdef other than Assembly-CSharp, it works.
While the error is happening, Test Runner won't read your test cases, so you can't use Test Runner.

I don't know if this is a bug or expected, and even if it is, I don't know if it's a VContainer's or a TestRunner’s, so I would appreciate it if you could check.

### Logs
StackOverflowException: The requested operation caused a stack overflow.
System.Globalization.CultureInfo.CreateCulture (System.String name, System.Boolean reference) (at <eae584ce26bc40229c1b1aa476bfa589>:0)
System.Reflection.AssemblyName.FillName (Mono.MonoAssemblyName* native, System.String codeBase, System.Boolean addVersion, System.Boolean addPublickey, System.Boolean defaultToken, System.Boolean assemblyRef) (at <eae584ce26bc40229c1b1aa476bfa589>:0)
System.Reflection.Assembly.GetReferencedAssemblies (System.Reflection.Assembly module) (at <eae584ce26bc40229c1b1aa476bfa589>:0)
[...]

PR #168, opened 2021-03-28
Make sure Assembly-CSharp is not a target of CodeGen

hadashiA:
Currently, any assembly that references `VContainer.EnableCodeGen` is a target of CodeGen optimization.

I noticed that Assembly-CSharp is targeted by default.
Since this is unnecessary and may lead to unintended behavior, we will disable CodeGen by default.


hadashiA:
#167

PR #200, opened 2021-04-24
 Support dynamic registration from open generic type

hadashiA:
#301 

If the open generic type is registered, the closed generic type will be dynamically assembled at resolve time.

```csharp
builder.Register(typeof(IFoo<,>), typeof(Foo<,>))
```

📝  Usage:
https://github.com/Cysharp/MessagePipe/blob/master/src/MessagePipe/ServiceCollectionExtensions.cs#L43-L102


igor84:
Since this approach is only making instances of the closed generic types using reflection it means it won't work with IL2CPP, right? It will only work if code generation is used since it will then generate the code that actually uses those types so they won't be missing in runtime.

hadashiA:
I decided to adopt this one.  #367 

MakeGenericType seems to work in recent Unity. See also the comments.

Issue #217, opened 2021-05-05
Reverse injection order


longbombus:
ReflectionInjector now injects Methods, Properties then Fields.
Would it be safer to reverse order? In that case Methods are injected last and can refer to other injectables. By the way Properties setters will availabled to refer to injectable Fields.

As I understand Properties/Fields injection means optional dependency, also Method injection is like constructor mean entry point of initialization. So in injected method we can figure out which optional dependencies are resolved.

PR #227, opened 2021-05-13
Injection order is reversed

longbombus:
https://github.com/hadashiA/VContainer/issues/217


longbombus:
I find out that tests not passing if codegen is enabled. I just copy-pasted generation in reverse order in InjectionILGenerator in new commit.

hadashiA:
👍   Thanks for adding the tests

PR #252, opened 2021-06-04
Proofread unitask.mdx

JesseTG:


hadashiA:
Thanks again.
Maybe this is more helpful and the sentence looks more natural.
I'd like to adopt it.

JesseTG:
Happy to help. VContainer is great, and the documentation is already pretty good (especially because your English is better than my Japanese). I just wanted to give it that extra little push that it deserves.

PR #264, opened 2021-06-17
Proofread method-injection.mdx

JesseTG:
- This time I've added some new information about fitting `[Inject]` in with the Unity life cycle.

### DIFF
--- a/website/docs/resolving/method-injection.mdx
+++ b/website/docs/resolving/method-injection.mdx
@@ -3,9 +3,7 @@ id: method-injection
 title: Method Injection
 ---
 
-If constructor injection is not available, use method injection.
-
-Typically this is for MonoBehaviour.
+If you need to inject dependencies into a type but can't do so with constructors, consider using method injection instead. This is most useful for `MonoBehaviour`s but can be done with any class, including test cases within the [Unity Test Framework](https://docs.unity3d.com/Packages/com.unity.test-framework@latest).
 
 ```csharp
 public class SomeBehaviour : MonoBehaviour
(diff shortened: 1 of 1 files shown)


website/docs/resolving/method-injection.mdx
```diff
@@ -20,14 +18,14 @@ public class SomeBehaviour : MonoBehaviour
 }
 ```
 
-:::tip Recommendation
-Consider whether injection to MonoBehaviour is necessary.
-In a code base where domain logic and presentation are well decoupled, MonoBehaviour should act as a View component.
+The `[Inject]`-annotated method can have any name and any accessibility level.
 
-In my opinion, View components should only be responsible for rendering and should be flexible.
+If you inject dependencies into `MonoBehaviour`s, you must be mindful of when the `[Inject]`-annotated method is called with respect to [Unity's object lifecycle](https://docs.unity3d.com/Manual/ExecutionOrder). If your [`Awake()`](https://docs.unity3d.com/ScriptReference/MonoBehaviour.Awake), [`Start()`](https://docs.unity3d.com/ScriptReference/MonoBehaviour.Start), or [`OnEnable()`](https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnEnable) methods need any dependencies from VContainer, consider moving their logic into your injection method.
```
hadashiA, line 23: This is true. However, I do not want to recommend writing logic inside Inject methods.

Actually, Im working on getting the Inject method to execute earlier than Awake. :p
( Perhaps that would be preferable.

For example, the following
https://github.com/hadashiA/VContainer/blob/master/VContainer/Assets/VContainer/Runtime/Unity/PrefabComponentRegistration.cs#L48

So, I'd like to weaken the wording of this, or put it a bit further down in `:::tip ~ :::`. how about this ?

hadashiA:
I've made some comments, but overall I think you've made the information easier to understand. Thanks.

website/docs/resolving/method-injection.mdx
JesseTG: I thought that `Awake()` methods were called as soon as the owning `GameObject` is instantiated, regardless of whether or not the `GameObject` itself is enabled?

JesseTG: What would you suggest instead of moving initialization to `[Inject]` methods (both for right now and for when you implement the changes you're working on)?

JesseTG: Oh, and this might be important; the reason I suggest moving initialization to `[Inject]` is that in my own project I spawn almost all prefabs using Addressables, so I can't use `IObjectResolver.Instantiate` (or whatever the extension method is called).

hadashiA: >  I thought that Awake() methods were called as soon as the owning GameObject is instantiated, regardless of whether or not the GameObject itself is enabled?

There is documentation on this:
>  Awake: This function is always called before any Start functions and also just after a prefab
is instantiated. (If a GameObject is inactive during start up Awake is not called until it is made active.)

https://docs.unity3d.com/Manual/ExecutionOrder.html
 

>  What would you suggest instead of moving initialization to [Inject] methods (both for right now and for when you implement the changes you're working on)?

I think that `Start()` is fine for most processes.
( As you said, Awake / OnEnable may be a timing issue. However, I think Start can be controlled to be called later than Inject.

In fact, I don't see a problem with doing a lot of things in the Inject method.

On the other hand, I think the constructor of an object should avoid causing side effects to other objects or executing operations that cause exceptions.

For the [Inject] method, I prefer to keep it to a role almost similar to that of a constructor.

hadashiA:
I have confirmed that you have changed the order.
I will merge them.

website/docs/resolving/method-injection.mdx
JesseTG: > 
> 
> > I thought that Awake() methods were called as soon as the owning GameObject is instantiated, regardless of whether or not the GameObject itself is enabled?
> 
> There is documentation on this:
> 
> > Awake: This function is always called before any Start functions and also just after a prefab
> > is instantiated. (If a GameObject is inactive during start up Awake is not called until it is made active.)
> 
> https://docs.unity3d.com/Manual/ExecutionOrder.html

Neat. I might submit another revision that explicitly refers to this, then.

> > What would you suggest instead of moving initialization to [Inject] methods (both for right now and for when you implement the changes you're working on)?
> 
> I think that `Start()` is fine for most processes.
> ( As you said, Awake / OnEnable may be a timing issue. However, I think Start can be controlled to be called later than Inject.
> 
> In fact, I don't see a problem with doing a lot of things in the Inject method.

Neither do I, which is why I asked. But it's nice to know that it's not the only option.

> On the other hand, I think the constructor of an object should avoid causing side effects to other objects or executing operations that cause exceptions.
> 
> For the [Inject] method, I prefer to keep it to a role almost similar to that of a constructor.

True, but I consider that more of a (strong) guideline. Sometimes you have to bend the rules so they don't break.

PR #269, opened 2021-06-23
Proofread container-api.mdx

JesseTG:
- Be more explicit about the automatic registration of IObjectResolver and LifetimeScope
- Advise on writing extension methods
- Add more details on what the code sample does

### DIFF
--- a/website/docs/resolving/container-api.mdx
+++ b/website/docs/resolving/container-api.mdx
@@ -38,10 +38,10 @@ class ClassA
 }
 ```
 
-If you regularly use `IObjectResolver` in a certain pattern, consider writing an [extension method](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods) for it. In fact, almost every `IObjectResolver` API is an extension method (including in the preceding code sample).
+If you regularly use `IObjectResolver.Inject` in a certain pattern, consider writing an [extension method](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods) for it. In fact, almost every `IObjectResolver` API is an extension method (including in the preceding code sample).
 
 :::caution
-Only use `IObjectResolver` explicitly if the other supported injection techniques don't fit your needs, as it's slow and not optimized by the [IL code generator](../optimization/codegen). Even VContainer callbacks that expose `IObjectResolver` (e.g. [build callbacks](../registering/register-callbacks) or [certain factory functions](../registering/register-factory#register-func-factory-that-requires-container-dependencies-and-runtime-parameters)) should minimize its use.
+Only use `IObjectResolver.Resolve` explicitly if the other supported injection techniques don't fit your needs, as it's slow and not optimized by the [IL code generator](../optimization/codegen). Even VContainer callbacks that expose `IObjectResolver` (e.g. [build callbacks](../registering/register-callbacks) or [certain factory functions](../registering/register-factory#register-func-factory-that-requires-container-dependencies-and-runtime-parameters)) should minimize its use.
 
 `LifetimeScope` (which VContainer also registers automatically) has an `IObjectResolver` reference through its `Container` property.
 
--- a/website/docs/resolving/container-api.mdx
+++ b/website/docs/resolving/container-api.mdx
@@ -32,6 +32,8 @@ class ClassA
         // other means (e.g. procedurally or from Addressables)
         // then consider using InjectGameObject.
         var object1 = container.Instantiate(prefab);
+
+        // There are also overloads that mimic Object.Instantiate.
         var object2 = container.Instantiate(prefab, parent);
         var object3 = container.Instantiate(prefab, position, rotation, parent);
     }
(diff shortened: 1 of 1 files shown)


website/docs/resolving/container-api.mdx
```diff
@@ -11,36 +11,51 @@ class ClassA
...
     }
 }
 ```
 
+If you regularly use `IObjectResolver.Inject` in a certain pattern, consider writing an [extension method](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods) for it. In fact, almost every `IObjectResolver` API is an extension method (including in the preceding code sample).
+
 :::caution
-These methods have poor performance (especially if CodeGen is not enabled). Use only if normal DI patterns are not applicable.
-:::
+Only use `IObjectResolver.Resolve` explicitly if the other supported injection techniques don't fit your needs, as it's slow and not optimized by the [IL code generator](../optimization/codegen). Even VContainer callbacks that expose `IObjectResolver` (e.g. [build callbacks](../registering/register-callbacks) or [certain factory functions](../registering/register-factory#register-func-factory-that-requires-container-dependencies-and-runtime-parameters)) should minimize its use.
```
hadashiA, line 44: >  it's slow and not optimized by the [IL code generator](../optimization/codegen).

About this point. My original documentation may have been misleading. ><
Let me describe the details.

- The optimization by IL code gen is actually done even when using IObjectResolver directly.
-  If we are *not* using IL code gen, `IObjectResolver.Inject` , `.Instantiate`, etc. are slower because the type checking by reflection happens at that point. 
    - However, it will be cached.
- `.InjectGameObject` etc are simply slow because it searches for a MonoBehaviour.
- `IObjectResolver.Resolve`, the performance is probably due to the direct call.

JesseTG: > * The optimization by IL code gen is actually done even when using IObjectResolver directly.
> * If we are _not_ using IL code gen, `IObjectResolver.Inject` , `.Instantiate`, etc. are slower because the type checking by reflection happens at that point.
>   * However, it will be cached.

Should I soften the language warning about the performance implications, then?

> * `.InjectGameObject` etc are simply slow because it searches for a MonoBehaviour.

Have you considered introducing an `IInjectableBehaviour` interface that serves as an optional target for [`MonoBehaviour.GetComponentsInChildren`](https://docs.unity3d.com/2021.1/Documentation/ScriptReference/Component.GetComponentsInChildren.html)? I suspect that it would be faster to let the engine find the relevant `MonoBehaviour`s than do to it in C#, although only a benchmark can say that for sure.

> * `IObjectResolver.Resolve`, the performance is probably due to the direct call.

Direct call as opposed to what? I'm not sure I understand.

---

On an unrelated note, would you mind looking at [this question of mine](https://github.com/hadashiA/VContainer/discussions/267) when you get a chance? I ask because I don't seem to get notifications for discussions, so I'm not sure if other people do either. Also that question is something I've been stuck on for months.

hadashiA: > Should I soften the language warning about the performance implications, then?

The only thing that is a little different is that IL CodeGen is not applied. The rest is fine as it is.

>  Have you considered introducing an `IInjectableBehaviour` interface that serves as an optional target for MonoBehaviour.GetComponentsInChildren? 

There doesn't seem to be much advantage in having the user pay for the implementation of the interface.

DI is, after all, the magic of reflection, so the cost of checking for types cannot be reduced.

It also seems simpler if the class definition code is less dependent on the DI. (Of course, we can't do this perfectly, but...)

>  On an unrelated note, would you mind looking at this question of mine when you get a chance? 

Oh, I missed that. I'll check later  #267  .

I haven't had much time lately. I'm sorry.
Thanks for your contribution !
#267

JesseTG: > The only thing that is a little different is that IL CodeGen is not applied. The rest is fine as it is.

Okay, so I'll frame it in terms of usability rather than performance.

> There doesn't seem to be much advantage in having the user pay for the implementation of the interface.

The idea isn't to check *all* `MonoBehaviour`s for dependencies; the idea is to *only* check `MonoBehaviour`s that might need them. Given that what I'm proposing can be done without modifying existing VContainer code, I might just prepare a benchmark in my project and let you know how it goes.

> DI is, after all, the magic of reflection, so the cost of checking for types cannot be reduced.

Fair enough. I know that according to [this talk](https://www.youtube.com/watch?v=8hru629dkRY) Zenject was never a dominant cost for Pokemon Go, a triple-A game released six years ago (and therefore with six fewer years of performance and usability improvements). It stands to reason that VContainer, fast as it is, would be even less of a bottleneck.

> It also seems simpler if the class definition code is less dependent on the DI. (Of course, we can't do this perfectly, but...)

I take the opposite philosophy; I use DI and `[Inject]` very generously. I tried passing around the majority of my game's dependencies with event arguments and `[Serializable]` fields, but that actually made things more complicated:

- When I used event arguments to pass dependencies, I still had to know who was sending the event. I also had to put all of my dependencies in an enormous `IDependencyProfile` interface.
- When I used `[Serializable]` fields, it complicated my use of Addressables and platform-specific assets.
- Remember, I'm not just passing dependencies to `MonoBehaviour`s.
- Some of my dependencies even depend on each other (although not in a cyclical way). And sometimes whether or not that's true varies by platform. For example, at least one platform I'm using requires providing user information (user ID, logged-in status, etc.) to use the storage API.

> I haven't had much time lately. I'm sorry.
> Thanks for your contribution !

No problem, happy to help! Thank you for this great package; it's actually the first DI library I've ever used, and now that I "get" it I actually enjoy using it.

JesseTG (commented):
How's this? Is this information accurate?

hadashiA:
Thanks for the fix and the comments. 
I will merge this PR!

https://github.com/hadashiA/VContainer/pull/269#discussion_r659198300 is a bit of a different topic, so I guess we can talk about it in #267 or something .

PR #279, opened 2021-07-10
Fix allocation on tickables

adarapata:
When using Tickable, GC Allocation of 40 bytes per frame was occurring. 
`IEnumerable<T>.GetEnumerator()` causes an implicit cast, I think.

To solve this problem, I modified the LoopItem to explicitly hold the List inside.

I'd like you to see if it looks OK. 🙏

### DIFF
--- a/VContainer/Assets/VContainer/Runtime/Unity/PlayerLoopItem.cs
+++ b/VContainer/Assets/VContainer/Runtime/Unity/PlayerLoopItem.cs
@@ -186,7 +186,7 @@ namespace VContainer.Unity
 
     sealed class PostFixedTickableLoopItem : IPlayerLoopItem, IDisposable
     {
-        readonly IEnumerable<IPostFixedTickable> entries;
+        readonly List<IPostFixedTickable> entries;
         readonly EntryPointExceptionHandler exceptionHandler;
         bool disposed;
 
@@ -194,7 +194,7 @@ namespace VContainer.Unity
             IEnumerable<IPostFixedTickable> entries,
             EntryPointExceptionHandler exceptionHandler)
         {
-            this.entries = entries;
+            this.entries = entries.ToList();
             this.exceptionHandler = exceptionHandler;
         }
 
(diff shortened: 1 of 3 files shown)


VContainer/Assets/VContainer/Runtime/Unity/PlayerLoopItem.cs
```diff
@@ -149,15 +150,15 @@ public bool MoveNext()
...
+        readonly List<IFixedTickable> entries;
         readonly EntryPointExceptionHandler exceptionHandler;
         bool disposed;
 
         public FixedTickableLoopItem(
             IEnumerable<IFixedTickable> entries,
             EntryPointExceptionHandler exceptionHandler)
         {
-            this.entries = entries;
+            this.entries = entries.ToList();
```
hadashiA, line 161: I think it would be better if `ToList()` was removed.
( Perhaps the copy occurs even if the concrete type is already a List.

This class is internal and has no use case other than being passed an `IReadOnlyList<T>`.
In this case, why not keep `IReadOnlyList<T>` as an argument, and use `for` and `[]` operator instead of `foreach` ?
( `for ...` statement  is exactly the same as the implementation of List<>.GetEnumerator. )

VContainer/Assets/VContainer/Tests/Unity/PlayerLoopItemTest.cs
```diff
@@ -0,0 +1,100 @@
...
+            public void MoveNextWithoutAllocation()
+            {
+                var list = new List<IPostLateTickable> { new Ticker(), new Ticker() };
+                var exceptionHandler = new EntryPointExceptionHandler(exception => { });
+                var tickableLoopItem = new PostLateTickableLoopItem(list, exceptionHandler);
+                
+                Assert.That(() =>
+                {
+                    tickableLoopItem.MoveNext();
+                }, Is.Not.AllocatingGCMemory());
```
hadashiA, line 64: 👍

VContainer/Assets/VContainer/Runtime/Unity/PlayerLoopItem.cs
adarapata: >In this case, why not keep IReadOnlyList<T> as an argument, and use for and [] operator instead of foreach ?

I like that!

adarapata:
Thanks for the review!
I've responded.

hadashiA:
This makes sense. Thanks!

Issue #271, opened 2021-06-23
Execution order for entry points


mnarimani:
Is it possible to specify execution order for non-MonoBehaviour classes that are created through `RegisterEntryPoint`?

hadashiA:
Currently, this feature is not available.

I feel that controlling the order at the class level is a bit complicated.

There are several timings available, so you can separate the groups.
https://vcontainer.hadashikick.jp/integrations/entrypoint#available-interfaces

4nonym0us:
The feature is important for Unity-oriented DI containers because there are cases where having only couple timings is not enough to cover the requirement:
* Unity provides API (`DefaultExecutionOrderAttribute`, `MonoImporter.SetExecutionOrder()`) and allows to set the execution order in Project Settings as well.
* Zenject provides [API](https://github.com/modesttree/Zenject#update--initialization-order) to control the execution order of ITickables and IInitializables as well as object disposal order (note that objects are disposed in a reversed order).

In some basic cases having just XEvent and PostXEvent is sufficient, but it's not unusual that a service\dependency must be `IInitialized` earlier/later then others. For example,
* It's a common desire to have core/infrastructure services (error handling/logging/telemetry/etc) initialized early and disposed last.
* Some services/dependencies need to be initialized before others (i.e. game saving/loading system needs to get initialized early to restore the game state).
Having just 2 timings might not be enough for sophisticated use-cases with complex initialization or where it's necessary to control the execution order of scripts within the same timing.

4nonym0us:
I've created a PR with a basic implementation.

Public API for controlling execution order is [here](https://github.com/hadashiA/VContainer/pull/287/files#diff-d2ed92bd2edb71ac8b36cbc73a899a7b1e4cad629a6c8a7113a25b02d140df76R109-R114). Corresponding unit test is located [here](https://github.com/hadashiA/VContainer/pull/287/files#diff-55baee67399301139a65ee580dc18208c3f6bc2dcb9b3140af3f6636a6ddf5e5R390).
Execution order number works just like Unity's or Zenject's ExecutionOrder: the lower the number, the earlier player loop timing is executed. Disposal of objects which implement `IDisposable` is performed in reversed order (similar to Zenject).

Let me know what you think about the feature and the implementation in particular.
I can add new API to docs if you think that PR is worth merging.

Artein:
[...]

PR #287, opened 2021-07-22
Add API to define the execution order of player loop timings.

4nonym0us:
Implements #271


hadashiA:
Sorry, but we could not adopt this specification and would like to close this case.

PR #294, opened 2021-08-10
Proofread property-field-injection.mdx

JesseTG:
- Removed the claim that optional dependencies are supported (since the code does not reflect this)


hadashiA (approved):
👍

Issue #298, opened 2021-08-11
GetRuntimeParent doesn't always work for parents specified in inspector


revolt3r:
It seems like the code is looking for root objects in the scene, but that won't work if you instantiate a Scope by code from a prefab or moved the scope to DontDestroyOnLoad..
Why don't you use `Object.FindObjectsOfType`?

hadashiA:
#299 merged !

PR #299, opened 2021-08-11
Fix Scope Parent searching when the parent is marked as DontDestroyOnLoad

revolt3r:
Fixes this issue: https://github.com/hadashiA/VContainer/issues/298


hadashiA:
Thanks for the pull request !

If this behavior is the same as FindObjectOfType, I'd love to merge it.
https://github.com/hadashiA/VContainer/pull/299/files#diff-e177703e4e2abf22e9aa65a552e0747392037e88ad30b3c15cecedbf6b2256e4L91

Could you please check it?

PR #300, opened 2021-08-11
Add inline API documentation

JesseTG:
This isn't ready yet, but I'd like your feedback as I work my way through the public APIs.


hadashiA:
I am sorry. While I think this was a useful effort, my review became a bottleneck and the content has become outdated.
I believe it will likely be necessary to restart from a different branch, so please allow me to close this.
Thank you for your contribution.

Issue #288, opened 2021-07-23
Resolving collections doesn't work with parent scopes


pnarimani:
Given two scopes "ParentScope" and "ChildScope", multiple registrations of an interface in "ParentScope" will not be available in objects that are registered in "ChildScope".
For example:

```
public interface ISample{}
public class A : ISample{}
public class B : ISample{}


// Parent scope is the Root scope which is defined in VContainer Settings
public class ParentScope : LifetimeScope 
{
        public A a;
        public B b;
        protected override void Configure(IContainerBuilder builder)
        {
             builder.RegisterInstance(a).AsImplementedInterfaces();
             builder.RegisterInstance(b).AsImplementedInterfaces();
        }
}

// Child is set to use ParentScope as parent in the inspector
public class ChildScope : LifetimeScope 
{
      protected override void Configure(IContainerBuilder builder)
      {
           builder.RegisterEntryPoint<User>();
      }
}

public class User : IInitializable
{
    public User(IReadOnlyList<ISample> samples)
    {
           // Here samples length is 0 even though A and B are registered as ISample in ParentScope
     }

     public void Initialize(){}
}
```

hadashiA:
Fixed in #321
Any feedback is welcome

PR #305, opened 2021-09-08
Fix collection resolve on parent

adarapata:
refs #288 

- Fixed collection resolution to follow the parent container.
- Added validation so that only lifecycle interfaces such as IInitializable are not traced as exceptions

It works, but it's a little less beautiful, and I want you to see it.

### DIFF
--- a/VContainer/Assets/VContainer/Runtime/Annotations/AnnotationUtility.cs
+++ b/VContainer/Assets/VContainer/Runtime/Annotations/AnnotationUtility.cs
@@ -1,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace VContainer.Unity
+{
+    public static class AnnotationUtility
+    {
+        private static readonly List<Type> _list = new List<Type>
+        {
+#if VCONTAINER_UNITASK_INTEGRATION
+            typeof(IAsyncStartable),
+#endif
+            typeof(IInitializable),
+            typeof(IPostInitializable),
+            typeof(IStartable),
+            typeof(IPostStartable),
+            typeof(IFixedTickable),
+            typeof(IPostFixedTickable),
+            typeof(ITickable),
+            typeof(IPostTickable),
+            typeof(ILateTickable),
+            typeof(IPostLateTickable)
+        };
+
+        public static bool IsLifecycleAnnotation(Type type) => _list.Contains(type);
+    }
+}
--- a/VContainer/Assets/VContainer/Runtime/Container.cs
+++ b/VContainer/Assets/VContainer/Runtime/Container.cs
@@ -2,6 +2,7 @@
 using System.Collections.Concurrent;
 using System.Runtime.CompilerServices;
 using VContainer.Internal;
+using VContainer.Unity;
 
 namespace VContainer
 {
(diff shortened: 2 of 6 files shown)


VContainer/Assets/VContainer/Runtime/Annotations/AnnotationUtility.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace VContainer.Unity
+{
+    public static class AnnotationUtility
+    {
+        private static readonly List<Type> _list = new List<Type>
+        {
+#if VCONTAINER_UNITASK_INTEGRATION
+            typeof(IAsyncStartable),
+#endif
+            typeof(IInitializable),
```
hadashiA, line 13: If we give special treatment to `IInitializable` etc and so on, the behavior when the user uses `IEnumerable<IInitializable>` will be different.

My idea is to have a special type just for this.
For example, how about `ContainerLocalCollection<T>`?

```csharp
sealed class ContainerLocalCollection<T> : IReadOnlyList<T>
{
    public int Count => innerList.Count;
    public T this[int index] => innerList[index];

    public IEnumerator<T> GetEnumerator() => innerList.GetEnumerator();
    IEnumerator IEnumerable.GetEnumerator() => innerList.GetEnumerator();
    
    readonly IReadOnlyList<T> innerList;    
    
    public ContainerLocalCollection(IReadOnlyList<T> innerList)
    {
        this.innerList = innerList;
    }
}
```

Then use it where we want to narrow down the scope (e.g. `EntryPointDispatcher`).

```csharp
var initializables = container.Resolve<ContainerLocalCollection<IInitializable>>();
```

In a few places where we create collections internally, we need to branch if Registration.ImplementationType is ContainerLocalCollection<>. However, I think it will make the intent clearer.

VContainer/Assets/VContainer/Runtime/Container.cs
```diff
@@ -117,14 +118,42 @@ object CreateTrackedInstance(IRegistration registration)
...
+            
+            CollectionRegistration collectionRegistration = null;
             while (scope != null)
             {
                 if (scope.TryGetRegistration(type, out var registration))
                 {
-                    return registration;
+                    if (registration is CollectionRegistration collection)
+                    {
+                        var elementType = type.GetGenericArguments()[0];
```
hadashiA, line 129: This is where the allocation for `Type[]` comes in, and we want to avoid it if possible.

VContainer/Assets/VContainer/Runtime/Container.cs
```diff
@@ -117,14 +118,42 @@ object CreateTrackedInstance(IRegistration registration)
...
+                        if (collectionRegistration == null)
+                        {
+                            collectionRegistration = new CollectionRegistration(elementType);
+                        }
+                        if (AnnotationUtility.IsLifecycleAnnotation(elementType))
+                        {
+                            return registration;
+                        }
+                        
+                        foreach (var registration1 in collection)
```
hadashiA, line 139: Couldn't you shorten the processing of this method a bit by adding some functionality to CollectionRegistration itself?

Also, and this is my mistake, but CollectionRegistration's GetEnumerator() causes a boxing allocation.

hadashiA:
@adarapata It looks like it was stale, so I've taken it over here ( #305 ). Thanks for your work.

PR #321, opened 2021-11-21
Fix collection resolve on parent

hadashiA:
#288 

Forked from #305


VContainer/Assets/VContainer/Runtime/Internal/CollectionRegistration.cs
```diff
@@ -0,0 +1,89 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace VContainer.Internal
+{
+    sealed class CollectionRegistration : IRegistration, IEnumerable<IRegistration>
+    {
+        public struct Enumerator : IEnumerator<IRegistration>
```
hadashiA, line 9: Optimization for `CollectionRegistration.GetEnumerator()`

Issue #330, opened 2022-01-01
Failure on injecting a prefab with PrefabComponentProvider will change the prefab's active state in Editor


yanuararistya:
## Expected Behavior
The prefab should not have its `active` state changed when injection has failed.

## Current Behavior

If a prefab started with the `active` state set to `true`,
the prefab will change its `active` state to `false` when `Inject` fails to resolve a dependency.

## Possible Solution

Add a `try finally` block when calling `injector.Inject`, and change the `active` state back to true inside the `finally` block.

## Steps to Reproduce

1. Create a `FooPresenter` class inheriting from `MonoBehaviour`, and add an unresolvable dependency to this class.
2. Create a prefab, and add `FooPresenter` component to it. Make sure that the prefab's gameObject `active` is set to `true`.
3. Register `FooPresenter` prefab with `builder.RegisterComponentInNewPrefab(prefab, LifetimeScope.Transient);`
4. Try to resolve `FooPresenter`. VContainer will spawn the object, but there will be an exception. Check that the `FooPresenter` prefab will have its `active` state changed to `false`.

## Detailed Description
 `PrefabComponentProvider.SpawnInstance` will instantiate a new instance of the prefab, but it will set the `active` state of the prefab to `false` before doing so. (I believe this is done so that we can inject the dependencies to the gameObject before Awake is called on it). If there is a failure when resolving its dependencies, the `active` state of this prefab will be left as `false`.

yanuararistya:
I implemented a possible solution for this issue on this pull request.

yanuararistya:
Closing the issue as the fix has been merged.

PR #331, opened 2022-01-01
PrefabComponentProvider.SpawnInstance will change the spawned prefab's active state on failed injection

yanuararistya:
A proposed solution for the issue described on https://github.com/hadashiA/VContainer/issues/330

`PrefabComponentProvider` will change the prefab's `active` state to inactive if there is a failure when injecting the spawned gameObject. This happens because `PrefabComponentProvider.SpawnInstance` changes the prefab to inactive before spawning it, and if `injector.Inject` throws an exception the function execution is halted, and the prefab's `active` state is left to inactive.

In order to prevent, we can add a try-finally block when calling `injector.Inject` function, and add re-activation of the prefab inside the `finally` block.

I have also added a unit-test for this case.


hadashiA:
Looks like better. Thanks!

PR #335, opened 2022-01-16
Feature Suggestion: ability to turn off ScriptTemplateModifier

YuukiARIA:
ScriptTemplateModifier is a nice feature, however we'd like to turn off this because we are using our own code generator.

So I created VContainerEditorSettings to save whether ScriptTemplateModifier is enabled or disabled.
And I changed ScriptTemplateModifier to work only if VContainerEditorSettings.EnableScriptModifier is true.

VContainerEditorSettings creates SettingsProvider, so that it appears in Project Settings > VContainer like as follows.

VContainerEditorSettings is saved as ProjectSettings/VContainerEditorSettings.asset, so this is project scope and will be version controlled.


hadashiA:
@YuukiARIA 
Thanks for the contribution .
I think it's a good idea, but can't we use the already existing VContainerSettings ?
https://github.com/hadashiA/VContainer/blob/master/VContainer/Assets/VContainer/Runtime/Unity/VContainerSettings.cs

YuukiARIA:
@hadashiA 
Thank you for replying!

Yes, it's no problem to use VContainerSettings.
The reason for VContainerEditorSettings is that I thought it might better to separete editor-only settings and runtime settings.

If it's OK to include the script template modifier option, I will include it to VContainerSettings.

hadashiA:
Oh, I see ><
At this point, I think it's easier to understand if the Setting assets are grouped together.
Therefore, I would appreciate it if you could use VContainerSettings.

YuukiARIA:
@hadashiA 
Oh, that makes sense.
I changed to use VContainerSettings, and the patch became much more simple!

Note that I also changed the default behavior of ScriptTemplateModifier to be enabled for backward compatibility.

hadashiA (approved):
👍 Thank you for your response. Sorry for the late reply.

Issue #333, opened 2022-01-14
When Resolve is called multiple times, the parent collection is bound to the child container.


adarapata:
Hi 😄 
I'm reporting what may be a glitch.

varsion: 1.10.0

Example
```cs
        [Test]
        public void ResolveCollectionFromParentByContinuous()
        {
            var builder = new ContainerBuilder();
            builder.Register<I1, MultipleInterfaceServiceA>(Lifetime.Singleton);
            builder.Register<I1, MultipleInterfaceServiceB>(Lifetime.Singleton);
            var parentContainer = builder.Build();

            var childContainer = parentContainer.CreateScope(childBuilder =>
            {
                childBuilder.Register<I1, MultipleInterfaceServiceC>(Lifetime.Singleton);
                childBuilder.Register<I1, MultipleInterfaceServiceD>(Lifetime.Singleton);
            });
            
            var scopedService = childContainer.Resolve<IReadOnlyList<I1>>();
            var moreScopedService = childContainer.Resolve<IReadOnlyList<I1>>();
            Assert.That(moreScopedService.Count(), Is.EqualTo(4));
        }
```

result
```
VContainer.VContainerException : Conflict implementation type : Registration MultipleInterfaceServiceA ContractTypes=[VContainer.Tests.I1] Singleton VContainer.Internal.InstanceProvider
```

If you check the contents of `CollectionRegistration` when you call `Resolve` for the second time, you will see that even the objects of the parent container are registered

```
registration = {Registration} Registration I1[] ContractTypes=[System.Collections.Generic.IEnumerable`1[VContainer.Tests.I1], System.Collections.Generic.IReadOnlyList`1[VContainer.Tests.I1]] Transient CollectionRegistration VContainer.Tests.I1[] ContractTypes=[System.Collections.Gener…
 ImplementationType = {RuntimeType} VContainer.Tests.I1[]
 InterfaceTypes = {List<Type>} Count = 2
 Lifetime = {Lifetime} Transient
 Provider = {CollectionInstanceProvider} CollectionRegistration VContainer.Tests.I1[] ContractTypes=[System.Collections.Generic.IEnumerable`1[VContainer.Tests.I1], System.Collections.Generic.IReadOnlyList`1[VContainer.Tests.I1]] Transient
   [0] = {Registration} Registration MultipleInterfaceServiceC ContractTypes=[VContainer.Tests.I1] Singleton VContainer.Internal.InstanceProvider
[...]

PR #336, opened 2022-01-18
fix FindRegistration for collection case

adarapata:
refs
- #333 

Clone the instance when resolving the collection.


hadashiA:
Nice fix 😄  Thanks for your quick respond.

PR #347, opened 2022-02-11
Fix syntax issue in site docs

KonH:
A semicolon is missing at https://vcontainer.hadashikick.jp/registering/register-scriptable-object


hadashiA:
Thanks 👍

PR #361, opened 2022-03-29
Singular/Plural/Code typo.

ceceomer:
changes:
are -> is
have -> has
; ->


hadashiA:
👍  Makes sense, thanks ><

Issue #301, opened 2021-08-25
 Support dynamic registration from open generic type


hadashiA:
For example, I would like to be able to quickly register a generic like this.
https://github.com/Cysharp/MessagePipe/blob/master/src/MessagePipe/ServiceCollectionExtensions.cs#L43-L102

slimshader:
Hi, what is the status of this one?

voreshmag:
Hi, I am also interested in this feature

slimshader:
Bump

hadashiA:
#367 merged.

PR #367, opened 2022-04-17
Add open generic registration api and resolve feature

DenisPimenov:
Hi! I try to implement open generic feature #301. If you have time to look and tell me what can be changed or improved I can do it


DenisPimenov:
Hey! Can you tell me if there is any progress on this issue?

hadashiA:
@DenisPimenov  Sorry for the delay in replying.

First of all, I would like to thank you. I think this implementation is not bad. 

However, I would like to confirm two points.


https://github.com/hadashiA/VContainer/pull/200#issuecomment-1221527508

> Since this approach is only making instances of the closed generic types using reflection it means it won't work with IL2CPP, right? It will only work if code generation is used since it will then generate the code that actually uses those types so they won't be missing in runtime.

As pointed out here, in an IL2CPP environment, classes that are not statically referenced will probably be omitted.
Therefore, in reality, I think it may be less problematic to list the closed generic type registrations instead of using the open generic api.

If this problem cannot be solved, there is probably a problem.


Second,  in some use cases like [MessagePipe](https://github.com/Cysharp/MessagePipe), 
we want all `ISubscriber` in the project to be registered, even if they are not referenced anywhere.
For example, as far as I know, autofac has the API like this: https://autofac.readthedocs.io/en/latest/register/scanning.html

So, my I understand is that, there are two ways to support generic types.

1. Dynamic generic type making at resolve time
2. Assembly scanning at container initialization time

If either or both of these are to be implemented, API conflicts need to be avoided.

( However, both 1 and 2 probably have IL2CPP issues...

DenisPimenov:
@hadashiA I will try different builds with il2cpp and report the results in a few days.

slimshader:
Is this still happening?

hadashiA:
Perhaps, the full generic sharing in Unity 2022.1 allows the following to work:

```cs
        var t1 = typeof(GenericClass<>).MakeGenericType(typeof(int));
        var i1 = (GenericClass<int>)Activator.CreateInstance(t1);

        var t2 = typeof(GenericClass<>).MakeGenericType(typeof(ClassA));
        var i2 = (GenericClass<ClassA>)Activator.CreateInstance(t2);

        var t3 = typeof(GenericClass<>).MakeGenericType(typeof(StructA));
        var i3 = (GenericClass<StructA>)Activator.CreateInstance(t3);

        var t4 = typeof(GenericStruct<>).MakeGenericType(typeof(int));
        var i4 = (GenericStruct<int>)Activator.CreateInstance(t4);

        var t5 = typeof(GenericStruct<>).MakeGenericType(typeof(ClassA));
        var i5 = (GenericStruct<ClassA>)Activator.CreateInstance(t5);

        var t6 = typeof(GenericStruct<>).MakeGenericType(typeof(StructA));
        var i6 = (GenericStruct<StructA>)Activator.CreateInstance(t6);
```

So this PR should probably work fine, as long as it's Unity 2022.1 or later.

I'd like to refine the API a bit more and merge this functionality.

DenisPimenov:
Hi. It has been a long time since opening PR. I have synchronized with master, but the tests are not passing. I will fix them.

DenisPimenov:
Hi! I fixed all the tests and fixed singleton resolution when the service is configured with multiple implementation types.

hadashiA:
@DenisPimenov 
Oh. Sorry. I was going to merge, can you just fix the conflicts?

DenisPimenov:
@hadashiA I have synchronized with master, but it has compilation errors

DenisPimenov:
Can you fix them?

VContainer/Assets/VContainer/Runtime/Registry.cs
```diff
@@ -109,12 +109,30 @@ public bool TryGet(Type interfaceType, out Registration registration)
...
+                    registration = implementationRegistration.GetClosedRegistration(interfaceType, typeParameters);
+                    return true;
+                }
+            }
+
+            registration = null;
+            return false;
+        }
+
         public bool Exists(Type type) => hashTable.TryGet(type, out _);
```
ch4ron, line 136:
```suggestion
        public bool Exists(Type type)
        {
            if (type.IsConstructedGenericType)
            {
                type = RuntimeTypeCache.OpenGenericTypeOf(type);
            }
            return hashTable.TryGet(type, out _);
        } 
```

I tested out your implementation in my project and ran into an issue here.
Exists doesn't behave correctly and returns false when checking registered open generics.

DenisPimenov: Thanks! I forgot the scopes tests and missed that error. I'll add them and your fix as soon as the compilation errors in the master branch are fixed

hadashiA:
>  I have synchronized with master, but it has compilation errors

Oh . . I see. 
This is another issue, so I'll check over here after the merge.

VContainer/Assets/VContainer/Runtime/Registry.cs
mixedHans: When I resolve IEnumerable<OpenGenericInterface<string>> or IReadOnlyList<OpenGenericInterface<string>>, I get an empty list. I expect to return me a list for all classes that are registered as this open generic interface. 

Additionally in regard to that, I suggest to add tests for resolving multiple instances of an interface.

PR #404, opened 2022-08-03
LifetimeScope: added generic overloads for CreateChild()

pspgit:
Added generic overloads for 
LifetimeScope.CreateChild`<TScope>`()

This allows us to create child scopes of specific type like this:
`var customScope = _parentScope.CreateChild<CustomScope>()`

Now we can select this programmatically created CustomScope to be a parent for some other scene scope, when scene is loaded.

image

### DIFF
--- a/VContainer/Assets/VContainer/Runtime/Unity/LifetimeScope.cs
+++ b/VContainer/Assets/VContainer/Runtime/Unity/LifetimeScope.cs
@@ -230,15 +230,9 @@ namespace VContainer.Unity
             return child;
         }
 
-        public LifetimeScope CreateChildFromPrefab(LifetimeScope prefab, IInstaller installer = null)
-            => CreateChildFromPrefab<LifetimeScope>(prefab, installer);
-
         public TScope CreateChildFromPrefab<TScope>(TScope prefab, Action<IContainerBuilder> installation)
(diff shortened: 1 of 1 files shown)


VContainer/Assets/VContainer/Runtime/Unity/LifetimeScope.cs
```diff
@@ -221,8 +230,15 @@ public LifetimeScope CreateChildFromPrefab(LifetimeScope prefab, IInstaller inst
             return child;
         }
 
+        public LifetimeScope CreateChildFromPrefab(LifetimeScope prefab, IInstaller installer = null)
```
hadashiA, line 233: I didn't see the need for both non-generic and generic methods.

If there is a generic method, it should be callable, like CreateChildFromPrefab(lifetimeScope).

Can we unify only the generic method, like CreateChild()?

pspgit: I've removed redundant CreateChildFromPrefab() overloads. 

Can't do the same for CreateChild<TScope>() without breaking backwards compatibility. 
It doesn't have a parameter to infer TScope type automatically.

hadashiA:
Thanks for the fix :)

PR #406, opened 2022-08-18
Fix a issue with root lifetimescope running even though not play mode

hadashiA:
Perhaps a Unity 2022.1 issue?

I noticed that VContainerSettings.OnEnable runs when the Editor Refresh completes.
I would like to add a check for isPlaying as a simple solution


YegorStepanov:
Unity 2020.3.36f1. 

After each(?) rebuild, the root scope executes **Configure()** in the Editor.

Add `Debug.Log` to see it.

YegorStepanov:
Ye, each rebuild it executes **Root.Configure**

PR fixes it for me.

LGTM

hadashiA:
@YegorStepanov Thanks for checking :)

PR #413, opened 2022-08-20
Make root.autoRun completely lazy

YegorStepanov:
Logically after #412, but not actually dependent on it.

The RootScope is not lazy, it build container immediately even if `sceneScope.AutoRun=false`.

After #412 commit:
```
Scene.AutoRun=false:
  Awake
  Root.Configure
  Awake end
```

After PR:
```
Scene.AutoRun=false:
  Awake
  Awake end
```

### DIFF
--- a/VContainer/Assets/VContainer/Runtime/Unity/LifetimeScope.cs
+++ b/VContainer/Assets/VContainer/Runtime/Unity/LifetimeScope.cs
@@ -161,6 +161,13 @@ namespace VContainer.Unity
 
             if (Parent != null)
             {
+                if (VContainerSettings.Instance != null && Parent == VContainerSettings.Instance.RootLifetimeScope)
+                {
+                    if (Parent.Container == null)
+                        Parent.Build();
+                }
+
+                // ReSharper disable once PossibleNullReferenceException
                 Container = Parent.Container.CreateScope(builder =>
                 {
                     builder.ApplicationOrigin = this;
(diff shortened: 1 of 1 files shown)


YegorStepanov:
Added a check, that this a root scope.

It may not be entirely clear what I wrote, but the problem is the following:
```C#
Awake()
{
    Parent = GetRuntimeParent(); //it executes Root.Build on the first SceneScope even if SCENEScope.autoRun=false
    if (autoRun)
    {
         Build(); //instead, it should call Root.Build in the Build method
    }
}
```

YegorStepanov:
Rebased it onto master.

PR:
```
root.AutoRun=false scene.AutoRun=false
    scene:       None    (only it differs)
    empty_scene: None

root.AutoRun=true scene.AutoRun=false
    scene:       Root.Build()
    empty_scene: Root.Build()

root.AutoRun=false scene.AutoRun=true
    scene:       Root.Build() Scene.Build()
    empty_scene: None

root.AutoRun=true scene.AutoRun=true
    scene:       Root.Build() Scene.Build()
    empty_scene: Root.Build()
```

Master:
```
root.AutoRun=false scene.AutoRun=false
    scene:       Root.Build()    (only it differs)
    empty_scene: None

root.AutoRun=true scene.AutoRun=false
    scene:       Root.Build()
    empty_scene: Root.Build()

root.AutoRun=false scene.AutoRun=true
    scene:       Root.Build() Scene.Build()
    empty_scene: None

root.AutoRun=true scene.AutoRun=true
    scene:       Root.Build() Scene.Build()
    empty_scene: Root.Build()
```

Ping @hadashiA

hadashiA:
The behaviors you summarized seems fine. I will merge them.

PR #414, opened 2022-08-21
Make Instantiate create objects in scope scene

### LINKED ISSUE #376: .Instantiate() should create objects in its scene
Under the hood, `IObjectResolver.Instantiate` uses 
```C#
Instantiate(prefab);
```
Unity creates the objects in the current active scene.

I think it should create objects in scope scene. 

To do this, the code should looks like: 
```C#
if(prefab is GameObject)
{
    var instance = Instantiate<GameObject>(prefab, currentScope.transform);
    instance.transform.parent = null;
}
```

When the scope is RootScope, it should create the objects in the active scene

[Edit]
Hmm, usually the first scene is temporary. It will inconvenient if RootScope objects will be destroyed soon. It would be better to create Root objects with "DontDestroyOnLoad" tag.

YegorStepanov:
Closes #376 

Don't know your API vision, let it be a prototype.

There is another option via `SceneManager.MoveGameObjectToScene()`, but it executes `Awake/Start` on the random scene, and is harder to pass to the plain C# part of the VContainer (`object ApplicationOrigin` already used in project to pass from Unity-C#-Unity layer).

### DIFF
--- a/VContainer/Assets/VContainer/Runtime/Unity/ObjectResolverUnityExtensions.cs
+++ b/VContainer/Assets/VContainer/Runtime/Unity/ObjectResolverUnityExtensions.cs
@@ -33,7 +33,20 @@ namespace VContainer.Unity
 
         public static T Instantiate<T>(this IObjectResolver resolver, T prefab) where T : UnityEngine.Object
         {
-            var instance = UnityEngine.Object.Instantiate(prefab);
+            var scope = (LifetimeScope)resolver.ApplicationOrigin;
+
+            T instance;
+            if (scope.IsRoot)
+            {
+                instance = UnityEngine.Object.Instantiate(prefab);
+                UnityEngine.Object.DontDestroyOnLoad(instance);
+            }
+            else
+            {
+                instance = UnityEngine.Object.Instantiate(prefab, scope.transform);
+                ResetParent(instance);
+            }
+
             InjectUnityEngineObject(resolver, instance);
             return instance;
         }
@@ -53,7 +66,19 @@ namespace VContainer.Unity
             Quaternion rotation)
             where T : UnityEngine.Object
         {
-            var instance = UnityEngine.Object.Instantiate(prefab, position, rotation);
+            var scope = (LifetimeScope)resolver.ApplicationOrigin;
+
+            T instance;
+            if (scope.IsRoot)
+            {
+                instance = UnityEngine.Object.Instantiate(prefab, position, rotation);
+                UnityEngine.Object.DontDestroyOnLoad(instance);
+            }
+            else
+            {
+                instance = UnityEngine.Object.Instantiate(prefab, position, rotation, scope.transform);
+                ResetParent(instance);
+            }
             InjectUnityEngineObject(resolver, instance);
             return instance;
         }
@@ -77,6 +102,19 @@ namespace VContainer.Unity
                 resolver.InjectGameObject(gameObject);
             else
                 resolver.Inject(instance);
+        }
+
+        static void ResetParent<T>(T instance) where T : UnityEngine.Object
+        {
+            switch (instance)
+            {
+                case Component component:
+                    component.transform.parent = null;
+                    break;
+                case GameObject gameObject:
+                    gameObject.transform.parent = null;
+                    break;
+            }
         }
     }
 }
--- a/VContainer/Assets/VContainer/Runtime/Container.cs
+++ b/VContainer/Assets/VContainer/Runtime/Container.cs
@@ -8,6 +8,7 @@ namespace VContainer
 {
     public interface IObjectResolver : IDisposable
     {
+        object ApplicationOrigin { get; }
         DiagnosticsCollector Diagnostics { get; set; }
 
         /// <summary>
@@ -47,6 +48,7 @@ namespace VContainer
     {
         public IObjectResolver Root { get; }
         public IScopedObjectResolver Parent { get; }
+        public object ApplicationOrigin { get; }
         public DiagnosticsCollector Diagnostics { get; set; }
 
         readonly Registry registry;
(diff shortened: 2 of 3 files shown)


hadashiA:
Thanks for the PR.  I am positive about this feature, but a little hesitant about adding IObjectResolver.ApplicationOrigin.
I will think about it and If I can't come up with any other good ideas, I would like to merge them. Thanks.

YegorStepanov:
Agree with you.

There are few issues about object creation, maybe we need deprecate these extension methods and give a more powerful solution?

For example:
```C#
public interface IObjectCreator
{
    //empty game object in the scope scene
    GameObject InstantiateEmpty(string name); //or CreateGameObject()

    //like current extension methods
    T Instantiate<T>(T prefab) where T : Object;
    T Instantiate<T>(T prefab, Vector3 position, Quaternion rotation) where T : Object;
    T Instantiate<T>(T prefab, Vector3 position, Quaternion rotation, Transform parent) where T : Object;
    T Instantiate<T>(T prefab, Transform parent, bool worldPositionStays = false) where T : Object;

    //A few places inside VContainer will use it (NewGameObjectResolver/Creation, PrefabCreation, don't remember correct names)
    //Sometimes it's impossible to resolve GameObject: precache or hold the reference in the upper scope, but resolve it current scope
    //It can be used for the optimization: do not traversal GameObject with a big structure
    T InstantiateNoInject<T>(T prefab) where T : Object;
    T InstantiateNoInject<T>(T prefab, Vector3 position, Quaternion rotation) where T : Object;
    T InstantiateNoInject<T>(T prefab, Vector3 position, Quaternion rotation, Transform parent) where T : Object;
    T InstantiateNoInject<T>(T prefab, Transform parent, bool worldPositionStays = false) where T : Object;

    //#352 and #304 issues
    TObject InstantiateWithParameter<TObject, TArg1>(TObject prefab, TArg1 arg1) where TObject : Object;
    TObject InstantiateWithParameter<TObject, TArg1, TArg2>(TObject prefab, TArg1 arg1, TArg2 arg2) where TObject : Object;
    TObject InstantiateWithParameter<TObject, TArg1, TArg2, TArg3>(TObject prefab, TArg1 arg1, TArg2 arg2, TArg3 arg3) where TObject : Object;
    TObject InstantiateWithParameter<TObject, TArg1, TArg2, TArg3, TArg4>(TObject prefab, TArg1 arg1, TArg2 arg2, TArg3 arg3, TArg4 arg4) where TObject : Object;

    //likewise:
    T InstantiateWithParameter<T,...>(T prefab, Vector3 position, Quaternion rotation) where T : Object;
    T InstantiateWithParameter<T,...>(T prefab, Vector3 position, Quaternion rotation, Transform parent) where T : Object;
    T InstantiateWithParameter<T,...>(T prefab, Transform parent, bool worldPositionStays = false) where T : Object;

    //mb plain C# methods?
    T Create<T>();
    T CreateNoInject<T>(); //for the unification
    T CreateWithParameter<T, TArg>();
}
```

YegorStepanov:
Few notes about code:

1) `component.transform.parent = null` gives a warning for the UI:
```
Parent of RectTransform is being set with parent property. 
Consider using the SetParent method instead, with the worldPositionStays argument set to false. 
This will retain local orientation and scale rather than world orientation and scale, which can prevent common UI scaling issues.
```

2) We should use overloaded version of the method:
```C#
component.transform.SetParent(null, worldPositionStays: true/false);
```

hadashiA:
> There are few issues about object creation, maybe we need deprecate these extension methods and give a more powerful solution?

I'm actually not very positive about this.
( In my experience, if DIContainer becomes a helper that dynamically creates objects, people will abuse it. I don't like that very much... 

I would suggest that objects that are dynamically generated at runtime should first consider using Factory. You can also control parameters as follows

```csharp
// before
builder.RegisterFactory<Param1, FooComponent>(container =>
{
    var dependency = container.Resolve<Dependency>();
    return param1 =>
    {
            var instance = container.Instanciate()
            instance.Configure(param1);
            return instance;
    };
}, Lifetime.Scoped);
```

I think this is sufficient, however it is also somewhat poor.
According to your suggestion, this could also be written much shorter.

```csharp
// maybe after
builder.RegisterFactory<Param1, FooComponent>(container =>
{
    return param1 => container.InstantiateWithParameter<FooComponent>(param1);
}, ...);
```

Not bad, but I am cautious as to whether this is necessary.
Either way, I thought it was a separate issue from this issue.


> ` IObjectResolver.ApplicationOrigin`

How about keeping ApplicationOrigin only used by ContainerBuilder?
Why not have the `RegisterComponent*` family all reference a scene and have the `ComponentRegistrationBuilder` always reference a scene? I think this issue could be solved.

VContainer/Assets/VContainer/Runtime/Unity/ObjectResolverUnityExtensions.cs
```diff
@@ -33,7 +33,20 @@ void InjectGameObjectRecursive(GameObject current)
...
         public static T Instantiate<T>(this IObjectResolver resolver, T prefab) where T : UnityEngine.Object
         {
-            var instance = UnityEngine.Object.Instantiate(prefab);
+            var scope = (LifetimeScope)resolver.ApplicationOrigin;
+
+            T instance;
+            if (scope.IsRoot)
+            {
+                instance = UnityEngine.Object.Instantiate(prefab);
+                UnityEngine.Object.DontDestroyOnLoad(instance);
```
hadashiA, line 42: It is intuitive, but it is a slightly different matter and a minor breaking change.

Instead of doing this, why not make the API explicit in the Register declaration?

like this:
`RegisterComponentInNewPrefab(...).DontDestroyOnLoad()`

hadashiA:
I have decided to adopt this PR.
Because I recently had an experience debugging scopes. And I came to think that linking Containers and GameObjects would help.

VContainer/Assets/VContainer/Runtime/Unity/ObjectResolverUnityExtensions.cs
hadashiA: I have reconsidered and will adopt this code as it is nonsense to Instantiate the root scope to the first scene.

Issue #409, opened 2022-08-20
VContainer events not firing for Scoped/Transient


YegorStepanov:
<details>
<summary>Template</summary>

```C#
public class RootScope : LifetimeScope
{
    protected override void Configure(IContainerBuilder builder)
    {
        //root

        builder.RegisterEntryPoint<Entry>();
    }

    private class Entry : IStartable
    {
        public Entry(Component1 c) { }

        public void Start() { }
    }
}

public class Scene1Scope : LifetimeScope
{
    protected override void Configure(IContainerBuilder builder)
    {
        //scene1
        
        builder.RegisterEntryPoint<Entry>();
    }

    private class Entry : IStartable
    {
        private readonly LifetimeScope _scope;

        public Entry(LifetimeScope scope, Component1 c) => _scope = scope;

        public void Start()
        {
            using (LifetimeScope.EnqueueParent(_scope))
                SceneManager.LoadScene("scene2", LoadSceneMode.Additive);
        }
    }
}

public class Scene2Scope : LifetimeScope
{
    protected override void Configure(IContainerBuilder builder)
    {
        //scene2

        builder.RegisterEntryPoint<Entry>();
    }

    private class Entry : IStartable
    {
        public Entry(Component1 c) {}

        public void Start() { }
    }
}

public class Component1: IInitializable, IStartable, IDisposable
{
    private LifetimeScope _lifetimeScope;

    public Component1(LifetimeScope s)
    {
        _lifetimeScope = s;
        Debug.Log("ctor " + _lifetimeScope.GetType().Name);
    }

    public void Initialize() => Debug.Log("init " + _lifetimeScope.GetType().Name);
    public void Start() => Debug.Log("start " + _lifetimeScope.GetType().Name);
    public void Dispose() => Debug.Log("dispose " + _lifetimeScope.GetType().Name);
}
```
</details>

1\)
```C#
//root
builder.Register<Component1>(Lifetime.Scoped).AsSelf().AsImplementedInterfaces();
//scene1

//scene2
```

VContainer events are called only in the root:

Result:
```
ctor RootScope
init RootScope
ctor Scene1Scope
start RootScope
ctor Scene2Scope
dispose Scene2
dispose Scene1
dispose Root1
```
Missing:
```
init Scene1Scope
init Scene2Scope
start Scene1Scope
start Scene2Scope
```

Should be:
```
ctor RootScope
init RootScope
ctor Scene1Scope
-init Scene1Scope
start RootScope
[...]

PR #425, opened 2022-09-21
Fixed a bug that entry-point events in Scoped/Transient does not work

hadashiA:
#409


hadashiA:
It seems to be too much branching by Registration/Provider type. I will refactor this.

Issue #434, opened 2022-10-27
ECS (beta) com.unity.entities@1.0.0-exp.8 support


CodeWiesel:
The new version of com.unity.entities breaks version 1.11.1

Fixed it for me with a few simple code changes. 
Probably not backwards compatible.

Creating the issue for PR reference.

PR #435, opened 2022-10-27
#434 ECS (beta) com.unity.entities@1.0.0-exp.8 support

CodeWiesel:


hadashiA:
@CodeWiesel Thanks a lot !

Sorry, but there seems to be a package version in the CI. Could you please fix this as well? 🙏 
https://github.com/hadashiA/VContainer/blob/master/.github/workflows/test.yaml#L65

Also, can you please correct any changes to the Unity version, etc. on this page?
https://github.com/hadashiA/VContainer/blob/master/website/docs/integrations/ecs.mdx

Thanks.

CodeWiesel:
@hadashiA 

Not sure how to test test.yaml / run the tests locally.
I updated the requirements etc. best to my knowledge...

hadashiA:
Thanks for the fixing 🙏 
I would like to merge this one, though.
But I'm not sure if all ecs users want to move to 2022.2 immediately.

- 2022.2  is still beta now.
- ecs 1.0.0-exp.8 means that may be experimental ?

So to get this into VContainer now, I would like to support both versions with compiler flags, etc, or wait for the official release.

I would like to do this work when I am available. Please wait a little.

hadashiA:
Merged #439

PR #441, opened 2022-11-08
fix: Exception thrown when InjectGameObject is reentered

alvinsay:
## FIX
- `InvalidOperationException: Collection was modified; enumeration operation may not execute.` will be thrown when InjectGameObject is re-entered. Check the test case `ReenterInjectGameObject()` for details.


alvinsay:
@hadashiA 

Any feedback?

hadashiA (approved):
Thanks for finding and fixing the bug  👍

PR #442, opened 2022-11-08
update: compatible to "enter play mode"

joe2643:
Dispose the container when loading into screen,
so that the container can be fully clean up without reload domain.


VContainer/Assets/VContainer/Runtime/Unity/VContainerSettings.cs
```diff
@@ -53,6 +53,7 @@ public static void LoadInstanceFromPreloadAssets()
             var preloadAsset = UnityEditor.PlayerSettings.GetPreloadedAssets().FirstOrDefault(x => x is VContainerSettings);
             if (preloadAsset is VContainerSettings instance)
             {
+                instance.RootLifetimeScope.DisposeCore();
```
hadashiA, line 56: Thanks for the fix.

I wonder if we need to consider the case where Instance.RootLifetimeScope is null?

joe264326: Thank you for reviewing.
You are right, we need a null checking of the RootLifetimeScope.

PR #444, opened 2022-11-08
Handle the case where a MonoBehaviour is null

JesseTG:
The array or list returned by `GameObject.GetComponents` might have a `null` element if a `MonoBehaviour`'s type couldn't be found (e.g. if it was stripped). See [here](https://docs.unity3d.com/2022.1/Documentation/ScriptReference/GameObject.GetComponents.html) for details.


hadashiA:
I think this is a hard-to-find bug, so I appreciate it very much. Thanks.

JesseTG:
You're welcome, I'm happy to help. Keep up the good work.

PR #452, opened 2022-11-28
Fix: Instantiation of a MonoBehaviour doesn't inject the entire instantiated GameObject

alvinsay:
When we pass a component to the `Object.Instantiate`, the Unity actually instantiates the entire GameObject that the MonoBehaviour belongs to. The returned MonoBehaviour instance actually attaches to a newly spawned GameObject which requires injection (initialization) also.

This patch fixes the issue that `T Instantiate(IObjectResolver resolver, T prefab ...)` series functions don't inject the entire spawned GameObject if a non-GameObject T is provided.


alvinsay:
@hadashiA Do you have time to give a review?

hadashiA:
+1 Makes sense, Thanks!

PR #454, opened 2022-12-04
Make ParentReference initializable

IShix-g:
This change sets the initial value for `LifetimeScope.parentReference` and is useful when [Project root LifetimeScope](https://vcontainer.hadashikick.jp/ja/scoping/project-root-lifetimescope) is not available. By calling in Reset(), it is automatically set when the script is attached.

```C#
using VContainer.Unity;

public sealed class GameSceneLifetimeScope : LifetimeScope
{
    void Reset() => parentReference = ParentReference.Create<AppLifetimeScope>();
}
```


hadashiA:
This will be merged since the impact is minimal 👍

Issue #432, opened 2022-10-19
Parameter injection works only with methods for MonoBehaviour


nikolay-vrjam:
Here is a MonoBehaviour with single parameter
```
public class VenueTestStarterController : MonoBehaviour, IStartable
    {
        public enum StartType
        {
            Active,
            Passive
        }

        [Inject] private StartType _startType; // This does NOT work

        // [Inject]  // This Works
        // private void Inject(StartType sType)
        // {
        //     _startType = sType;
        // }

        public void Start()
        {
            Debug.Log($"starttype={_startType}");
        }
    }
```
and this is a scope configuration
```
    protected override void Configure(IContainerBuilder builder)
        {
            builder.RegisterComponentOnNewGameObject<VenueTestStarterController>(Lifetime.Scoped,
                nameof(VenueTestStarterController)).WithParameter(VenueTestStarterController.StartType.Passive);
            builder.RegisterBuildCallback(resolver =>
            {
                resolver.Resolve<VenueTestStarterController>();
            });
        }
```
Injection only works when it happens through method (commented in the first snipped), otherwise, it shows following error
```
VContainerException: No such registration of type: Test.VenueTestStarter.VenueTestStarterController+StartType
VContainer.ScopedContainer.FindRegistration (System.Type type) (at Library/PackageCache/jp.hadashikick.vcontainer@1.11.1/Runtime/Container.cs:181)
VContainer.ScopedContainer.Resolve (System.Type type) (at Library/PackageCache/jp.hadashikick.vcontainer@1.11.1/Runtime/Container.cs:72)
VContainer.Internal.ReflectionInjector.InjectFields (System.Object obj, VContainer.IObjectResolver resolver) (at Library/PackageCache/jp.hadashikick.vcontainer@1.11.1/Runtime/Internal/ReflectionInjector.cs:66)
VContainer.Internal.ReflectionInjector.Inject (System.Object instance, VContainer.IObjectResolver resolver, System.Collections.Generic.IReadOnlyList`1[T] parameters) (at Library/PackageCache/jp.hadashikick.vcontainer@1.11.1/Runtime/Internal/ReflectionInjector.cs:26)
VContainer.Unity.NewGameObjectProvider.SpawnInstance (VContainer.IObjectResolver resolver) (at Library/PackageCache/jp.hadashikick.vcontainer@1.11.1/Runtime/Unity/InstanceProviders/NewGameObjectProvider.cs:44)
[...]

PR #457, opened 2022-12-06
Fix withParameter not working

adarapata:
refs
- #432 

In MonoBehaviour, WithParameter was not referenced except for method injection, which has been corrected.


hadashiA:
@adarapata 
I would like to adopt this feature, but I would like to have the same behavior for SourceGenerator as well as Reflection.
Is it possible for you to modify it?
https://github.com/hadashiA/VContainer/blob/master/VContainer.SourceGenerator/VContainerSourceGenerator.cs

adarapata:
I'll try it!

piti6:
Possible duplication of #156, but since it is too old to apply and I completely have no knowledges of recent VContainer codes, I'll close it.

PR #458, opened 2022-12-06
Remove itself from Inspector's Parent

IShix-g:
Added implementation to exclude own reference from `ParentReference` in `LifetimeScope`. Please consider this if you would like to do so.

screenshot


hadashiA:
I think it is appropriate in every case I can think of.

Im not 100% sure if we should strictly prohibit relationships between members of the same class, but I would like to take the benefit of this fix.
Thanks.

IShix-g:
Sorry to bother you.
If I ever commit again, I will try to submit it along with the tests, etc.

PR #467, opened 2023-01-22
Add UnmanagedSystem support in ECS Integration

stenyin:
This pull request adds support for UnmanagedSystem in the ECS Integration feature of VContainer. The changes include new extension methods for ContainerBuilder that allow for easy registration of UnmanagedSystems from a specific World, as well as new classes for handling the creation and disposal of UnmanagedSystem references.

This update also includes the ability to add a system to a specific system group using IntoGroup<T> method.

This should be a useful addition for those using UnmanagedSystems in their projects, and will make it easier to manage the lifetimes of these systems in conjunction with VContainer.

I've tested the changes and it's working well. Please let me know if there's anything else needed.

Note: this feature requires the VCONTAINER_ECS_INTEGRATION_1_0 preprocessor symbol to be defined in order to use it.


VContainer/Assets/VContainer/Tests/Fixtures.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+using Unity.Entities;
```
hadashiA, line 2: Maybe  we need the `#if ` directive ?

hadashiA:
@stenyin  Very interesting and appreciated as I have not been able to catch up on the new features of the ecs. thanks for the PR.

I would like to merge, but it would be very helpful if you could confirm two points.

1. If `VCONTAINER_ECS_INTEGRATION_1_0` and `VCONTAINER_ECS_INTEGRATION` are undefined, does it work as it was?
2.  > Note: this feature requires the VCONTAINER_ECS_INTEGRATION_1_0 preprocessor symbol to be defined in order to use it.
    - Do we need to do this manually?   It was assumed that if the project had a newer version of com.unity.entities, this preprosessor would be added automatically.
        - スクリーンショット 2023-01-31 13 53 38   

I would be happy to have it confirmed. Thanks.

stenyin:
1. if `VCONTAINER_ECS_INTEGRATION_1_0` and `VCONTAINER_ECS_INTEGRATION` are undefined, it work as well.
2. Entities 1.0 is compatible with Unity version 2022.2.0b8 and later. Can we check unity version instead of  `VCONTAINER_ECS_INTEGRATION_1_0` preprocessor symbol?

VContainer/Assets/VContainer/Tests/Fixtures.cs
stenyin: This is my mistake. i had remove this line.

hadashiA:
Thanks for the checking.

> 2. Entities 1.0 is compatible with Unity version 2022.2.0b8 and later. Can we check unity version instead of VCONTAINER_ECS_INTEGRATION_1_0 preprocessor symbol?

Sounds like a good idea.
The management the preproseccor symbol myself is complicated.
However, if unity 2022.2 can use less than ecs 1.0, then the unity version check is incomplete. How about this?

stenyin:
Thank you for your response.

I have confirmed that versions lower than ecs 1.0 cannot be used normally in 2022.2, so I believe using the Unity version as a check is a feasible solution (unless consideration is needed for those using 2022.2.0a1-2022.2.0b7 versions).

hadashiA:
Ok,  Could you fix this PR to checking the Unity version?

Borningstar:
Are there any additional changes needed before this can be merged in @hadashiA?

stenyin:
Sorry I haven't replied in a while, but for work and personal reasons. I'll test it again.

hadashiA:
Sorry for late reply. Looks good for me. Thanks!

PR #468, opened 2023-01-27
Supports private member injection into base classes.

nishikawa-joro:
I want the ability to inject private members of base classes.
I would appreciate your kind consideration.


hadashiA:
👍  👍 
Thanks for the PR.
Please let me consider this for a moment.

PR #470, opened 2023-01-30
added non generic RegisterInstance

voreshmag:
Hi! I added some new non-generic instance registration methods. It's useful sometimes, in example if you don't know your types in compile time and you want to register them in container. For example you may have configuration files that are deserialized in runtime and you don't know the exact type of this configs.


edamiani (approved):

edamiani:
Straight to the point, there's not much to comment on here. :)

The only thing I wanted to comment on is that I wish it would be possible to create an elegant solution using just one function with params at the end of the parameters list, something like:

public static RegistrationBuilder RegisterInstance(
            this IContainerBuilder builder,
            object instance,
            params Type[] interfaceTypes)

But I understand that it would be messy doing all the needed checks inside of it.

hadashiA:
I think your suggestion is not bad.
But what's the difference with `RegisterInstance(objct).As(type)`?

The method chain is a bit confusing though.
It can be commonly applied to any Register* method family.

If we were to add more method arguments, I would have to do it for the other Register* methods as well, otherwise it would be inconsistent.

So unless there is a strong reason to do so, I'd like to close it.

voreshmag:
> I think your suggestion is not bad. But what's the difference with `RegisterInstance(objct).As(type)`?
> 
> The method chain is a bit confusing though. It can be commonly applied to any Register* method family.
> 
> If we were to add more method arguments, I would have to do it for the other Register* methods as well, otherwise it would be inconsistent.
> 
> So unless there is a strong reason to do so, I'd like to close it.

I need this methods because if you will use `RegisterInstance(objct).As(type)` then VContainer will register object as object's type and then as "type". So if you want to register multiple objects with the same type there would be a registration conflict between them.

Example:
```
ConfigBase configA; //implements configA
ConfigBase configB; //implements configB

RegisterInstance(configA).As(configA);
RegisterInstance(configB).As(configB); <= registration conflict because ConfigBase already registered
```

hadashiA:
This works. Conflicts of type interface are allowed.

```csharp
            builder.RegisterInstance(instance1).As(typeof(IInterface);
            builder.RegisterInstance(instance2).As(typeof(IInterface));
```

Not allowed are the following cases
- Lifetime.Singleton.
- Registering multiple instances of the same implementation type
- In fact, RegisterInstance is set as Lifetime.Singleton

This is a bit odd, but for the following reasons
- The act of registering multiple instances of the same concrete type violates the behavior implied by "Singleton"
- One of the goals of the VContainer design is to avoid DI abuse. It is not recommended to register instances of the same type directly in a DI container.

Perhaps adding a non-generic RegisterInstance is not consistent with your purpose.
I am closing this.

PR #481, opened 2023-03-02
Fix asmdef entities expression to (0.60, 999)

airtonmotoki:
Problem: not working with entities 0.51

Fix: asmdef entities expression from (0.20, 999) to (0.60, 999)


hadashiA:
👍

PR #482, opened 2023-03-02
Better LifetimeScope.EnqueueParent

kamyker:
When using LifetimeScope.EnqueueParent a lot, it sometimes incorrectly removes the parent.

Code that fails without this fix:
```
using(LifetimeScope.EnqueueParent(parentScope))
{
	await Addressables.LoadSceneAsync(sceneRef);
       // another scene starts its own using(LifetimeScope.EnqueueParent(otherScope)) {
} // Dispose here removes parent set by other scene
```


hadashiA:
👍 👍

Issue #363, opened 2022-03-31
[Feature] automatic addition all registered classes to link.xml


vangogih:
Hi @hadashiA,

Based on issue #362 I suggest to add all registered methods or fields to link.xml file for preserving them.

hadashiA:
In the meantime, I believe the problem has been resolved with #362

vangogih:
I have already wrote that feature for my project.
And I want to release the conception and get your opinion.

PR #493, opened 2023-03-28
Draft: Automatic addition all registered classes to link.xml

vangogih:
Hi @hadashiA,

I have mentioned (#363) that I want to add the auto-addition all registered classes in scopes to link.xml.
Here is proof of concept. I have already tested generation for `MonoBehaviours`, `ScriptableObject` classes and classes with `Serializable` attribute.

How does it works:
I made the same mechanism to turn on the feature as CodeGen. Via aditional asmdef `VContainer.EnableLinkGenerator`. You need to add that assebmly as dependency to your runtime asmdef to enable that feature. So enabling you can find in class `VContainerLinkerProcessor`. That class tries to find the assembly with `VContainer` and `VContainer.EnableLinkGenerator` dependency. And if they are exist It starts to analyze all types in picked asseblies.

The analyzer consists of:
1. Finding all lifetime scopes and create their instances
2. Finding all declared fields with `SerializedField` attribute or derived from `Component` or `ScriptableObject`.
3. Initializing all have found fields via `AddComponent`, `ScriptableObject.Create` and `Activator.CreateInstance`
4. Calling method `Cofigure` on `LifetimeScope` to register all dependencies which defined in lifetime scope
5. Geting all registered types from builder

Next, I have wrote a custom link.xml builder class (`VContainerXmlLinkBuilder`) which based om unity's `UnityEditor.Build.Pipeline.Utilities.LinkXmlGenerator` class, but more powerful. It allows to preserve methods, fields and properties without whole class preserving (as `LinkXmlGemerator` does). I think that preserving all members by default is so frustrating feature.

That feature works with Vcontainer as backbox and doesn't require to change any parts of runtime.

I want to get your comment about that idea before continue. Feel free to give feedback 😄 
If you approve that feature I will:
- Resolve all TODO's
- Write tests for all parts of generation
- Extend field initializing by adding support `SerializeReference` attribute and recurcive fields initialization for `Serializable` classes
- Add documentation to site

p.s. feature is draft, but you can test in the editor if you need.

### DIFF
--- a/VContainer/Assets/VContainer/Runtime/ContainerBuilder.cs
+++ b/VContainer/Assets/VContainer/Runtime/ContainerBuilder.cs
@@ -68,7 +68,7 @@ namespace VContainer
             }
         }
 
-        readonly List<RegistrationBuilder> registrationBuilders = new List<RegistrationBuilder>();
+        internal readonly List<RegistrationBuilder> registrationBuilders = new List<RegistrationBuilder>();
         List<Action<IObjectResolver>> buildCallbacks;
         DiagnosticsCollector diagnostics;
 
--- a/VContainer/Assets/VContainer/Runtime/Internal/TypeAnalyzer.cs
+++ b/VContainer/Assets/VContainer/Runtime/Internal/TypeAnalyzer.cs
@@ -206,6 +206,7 @@ namespace VContainer.Internal
             var injectMethods = default(List<InjectMethodInfo>);
             var injectFields = default(List<FieldInfo>);
             var injectProperties = default(List<PropertyInfo>);
+            var startType = type;
             while (type != null && type != typeof(object))
             {
                 // Methods, [Inject] Only
@@ -245,7 +246,7 @@ namespace VContainer.Internal
             }
 
             return new InjectTypeInfo(
-                type,
+                startType,
                 injectConstructor,
                 injectMethods,
                 injectFields,
--- a/VContainer/Assets/VContainer/Editor/LinkGenerator.meta
+++ b/VContainer/Assets/VContainer/Editor/LinkGenerator.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: 7dd4e2ab029cc2f469d42fbe5bff6951
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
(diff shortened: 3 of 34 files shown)


VContainer/Assets/VContainer/Runtime/Internal/TypeAnalyzer.cs
```diff
@@ -206,6 +206,7 @@ public static InjectTypeInfo Analyze(Type type)
             var injectMethods = default(List<InjectMethodInfo>);
             var injectFields = default(List<FieldInfo>);
             var injectProperties = default(List<PropertyInfo>);
+            var startType = type;
```
vangogih, line 209: Here is the bug (or misunderstanding) which I fixed. The thing is `type` after while while recursive circle became `System.Object` for all classes. It is because type rewrites in 244-245 line as BaseType.

VContainer/Assets/VContainer/Runtime/ContainerBuilder.cs
```diff
@@ -68,7 +68,7 @@ public DiagnosticsCollector Diagnostics
             }
         }
 
-        readonly List<RegistrationBuilder> registrationBuilders = new List<RegistrationBuilder>();
+        internal readonly List<RegistrationBuilder> registrationBuilders = new List<RegistrationBuilder>();
```
vangogih, line 71: To get registered types from scope after calling `Configure` method I had to make that field internal.
See: LinkGeneratorTypeHelper.GetTypesFromLifetimeScopesInAsseblies:99

vangogih (commented):
Left an important comments about changes in runtime assembly.

vangogih:
@hadashiA ping 😀

hadashiA:
@vangogih 
Sorry for the late reply. I've been busy with another OSS work.

Big thanks for your a lot of work.  I think a lot of people would find the stripping measure interesting.

However, due to several points, I'm currently considering not including this in the main repo.

- Currently, I think Roslyn analyzer is the leading tool for static analysis of source code.
- Your algorithm is interesting, but maintaining both SoruceGenerator and this seems a bit much work.
- VContaienr has SoruceGenerator. If we can derive the classes that are being used reliably, it might be possible to mark them [Perserve]. I prefer that approach.

vangogih:
> @vangogih Sorry for the late reply. I've been busy with another OSS work.
> 
> Big thanks for your a lot of work. I think a lot of people would find the stripping measure interesting.
> 
> However, due to several points, I'm currently considering not including this in the main repo.
> 
> * Currently, I think Roslyn analyzer is the leading tool for static analysis of source code.
> * Your algorithm is interesting, but maintaining both SoruceGenerator and this seems a bit much work.
> * VContaienr has SoruceGenerator. If we can derive the classes that are being used reliably, it might be possible to mark them [Perserve]. I prefer that approach.

Thank for your review and comment!

I think either the PR is quite outdated now. And we can implement same functionality other ways with SourceGenerator feature.

I'm going to close this PR, thanks again)

Issue #496, opened 2023-03-30
[Advice] Best practice for DI usage on objects which derive from each other 


bmgoncu:
Hi folks,

Me and our team were wondering on the best practice of doing DI on objects which derive from each other.
Let us give an example;

```
public class A : MonoBehaviour
{
     [Inject]
     private void Inject(Dep1, Dep2)
     { ... }
}
```
```
public class B : A
{
     [Inject]
     private void Inject(Dep3, Dep4)
     { ... }
}
```
currently this causes issues on 1.13  (thats not the issue btw), but how should we tackle a situation like this? Whats the best practice there?

The same can be asked for Non-MonoBehaviour classes;
```
public class A
{
     [Inject]
     public A(Dep1, Dep2)
     { ... }
}
```
```
public class B : A
{
     [Inject]
     public B(Dep1, Dep2, Dep3, Dep4) : base(Dep1, Dep2)
     { ... }
}
```
How should we tackle this?

Please advice

ooonush:
Hi! This problem occurs with inheritance, and yes, this bug appeared in version 1.13.1. Inherited methods are called several times for each subtype in the inheritance hierarchy (even if there is no override modifier). Hopefully this will be fixed in the next update.

Issue #486, opened 2023-03-06
Virtual generated methods (Mono.Cecil) not supported in IL2CPP


ooonush:
If a type has virtual methods generated using Mono.Cecil, the game crashes when injecting into that type in the IL2CPP build.

This happens in the foreach loop when methodInfo is a generated method. The game crashes when methodInfo.IsDefined(typeof(InjectAttribute), true) is called.

Here is an example from script 212 TypeAnalyser.cs:
```cs
foreach (var methodInfo in type.GetRuntimeMethods())
{
    // Here, if there is a generated method, the game crashes
    if (methodInfo.IsDefined(typeof(InjectAttribute), true))
    {
        // do...
    }
}
```
This is most likely due to a compilation error. When generating the IsDefined virtual method, it starts an infinite recursion because the inherit parameter = true. This can be fixed by not using IsDefined, but writing the attribute check method yourself. Zenject has done this, so there is no such error there.

hadashiA:
Probably resolved in 505.
Also, Mono.Cecil is deprecated and SouceGenerator is recommended, so we will Close.

PR #500, opened 2023-04-11
Fix the member multiple injection

ooonush:
Fixed multiple injection in the inheritance. Issue: #496 and also #486.

There remains one more unpleasant bug that I could not fix elegantly: https://github.com/hadashiA/VContainer/issues/501

### DIFF
--- a/VContainer/Assets/VContainer/Runtime/Internal/TypeAnalyzer.cs
+++ b/VContainer/Assets/VContainer/Runtime/Internal/TypeAnalyzer.cs
@@ -206,37 +206,35 @@ namespace VContainer.Internal
             var injectMethods = default(List<InjectMethodInfo>);
             var injectFields = default(List<FieldInfo>);
             var injectProperties = default(List<PropertyInfo>);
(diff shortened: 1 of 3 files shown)


hadashiA:
Thanks a lot.

Let me check a few things.

Also, I think the SourceGenerator side needs to be fixed as well,
I may possibly re-create the PR on my end with this as a reference. Please understand.

hadashiA:
@ooonush Hi. I have created #505 with considerable help from this PR.
I think I was able to fix the problems in #496 , #486 and #501.
Thanks a lot. I would appreciate it if you could let me know if there are any problems.

ooonush:
Great, thanks. And what do you think about using HashSet to check for duplicate MemberInfo's? You could create an IEqualityComparer that compares by member.GetBaseDefinition(). That should work faster than a full search.

hadashiA:
> do you think about using HashSet to check for duplicate MemberInfo's?

Depends on the number of Inject methods.
In most cases, there will likely be at most one Inject method, so we consider the method with the least overhead to be the best.

PR #516, opened 2023-05-13
Fix wrong method visibility

left-:
Minor fix for the first code example at `website/docs/about/what-is-vcontainer.mdx`


hadashiA:
👍

Issue #489, opened 2023-03-17
Delegate registration impossible for Value Types


slimshader:
Hi,

I need to register a type that has to be accessed from another object by its getter so I have to use delegate registration (resolve + getter access) but it has generic constraint of a "class". Is there a reason for it? Other registration types do not have this limitation.
```
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        public static RegistrationBuilder Register<TInterface>(
            this IContainerBuilder builder,
            Func<IObjectResolver, TInterface> implementationConfiguration,
            Lifetime lifetime)
            where TInterface : class
            => builder.Register(new FuncRegistrationBuilder(implementationConfiguration, typeof(TInterface), lifetime));
```

hadashiA:
Ah, ok.
That's right. There is no consistency in this, and it's probably a mistake. I'll fix it.

slimshader:
> 

Great, what is the eta on that?

PR #521, opened 2023-05-22
Support for Value Types delegate registration

slimshader:
Fix for https://github.com/hadashiA/VContainer/issues/489


AlonTalmi:
I get compile error when removing the constraint without casting the delegate like so:

```
[MethodImpl(MethodImplOptions.AggressiveInlining)]
public static RegistrationBuilder Register<TInterface>(
    this IContainerBuilder builder,
    Func<IObjectResolver, TInterface> implementationConfiguration,
    Lifetime lifetime)
{
    return builder.Register(new FuncRegistrationBuilder(Create, typeof(TInterface), lifetime));
    object Create(IObjectResolver resolver) => implementationConfiguration(resolver);
}
```

hadashiA:
👍

PR #526, opened 2023-05-29
feat: add `Type` arg builder to `ContainerBuilderUnityExtensions`

### LINKED ISSUE #525: `ContainerBuilderUnityExtensions` API need variant that accept `Type` as args
Besides generic param, an api accept `System.Type` is needed for dynamic load types from assemblies

SettingDust:
fix #525


hadashiA:
:+1: I have confirmed that existing interfaces are saved .  I would like to merge this.

However, not all of the coding style changes appealed to me. Oh well.

PR #527, opened 2023-06-05
UniTask binding support from method

AlonTalmi:
### What I did:
I added a generic 
AsyncFuncInstanceProvider<T> : IInstanceProvider
And
AsyncFuncRegistrationBuilder<T> : RegistrationBuilder

which bind UniTask<T> instead of just T

Then added 2 new extension methods:
RegisterAsync<TInterface> which is used to bind UniTask<TInterface> from a method
ResolveAsync<TInterface> which return UniTask<TInterface> instead of TInterface

I also added a tests class to prove it works

and lastly wrapped all in VCONTAINER_UNITASK_INTEGRATION define to avoid compilation error in projects without UniTask

### Why?
I'm not sure what the best way to tackle async injections in DI without tackling it somehow from the root which would make things unreasonably complex, but I thought maybe supporting UniTask injection from method (which is currently not supported because UniTask is a struct) and adding a ResolveAsync extension will be better the nothing. but @hadashiA, tell me what you think and if I might miss something.

### DIFF
--- a/VContainer.StandaloneTests/AsyncRegisterTest.cs
+++ b/VContainer.StandaloneTests/AsyncRegisterTest.cs
@@ -1,0 +1,44 @@
+#if VCONTAINER_UNITASK_INTEGRATION
+using System.Threading.Tasks;
+using Cysharp.Threading.Tasks;
+using NUnit.Framework;
+
+namespace VContainer.Tests
+{
+    public class AsyncRegisterTest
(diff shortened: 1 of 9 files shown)


AlonTalmi:
I realized changing the binding type also means other extension methods will be broken as well :/
I'm closing this until I'll make it properly

PR #559, opened 2023-10-03
Fixed injection on MonoBehaviours not working if the registration was made using base class identifier

pnarimani:
Unit tests `RegisterComponentInHierarchyUsingBaseClass` and `RegisterComponentInNewPrefabUsingBaseClass` highlight the issue.
Fixed `FindComponentProvider` and `PrefabComponentProvider` so they actually inject the type that is being found/instantiated instead of the base type.


hadashiA:
Thanks! 👍

PR #564, opened 2023-10-30
Increase visibility of RegistrationBuilder and InjectorCache to make VContainer more extensible.

pnarimani:
With members of `RegistrationBuilder` being `internal`, it's very difficult to create your own project-specific registrations. 

I increased visiblity of members inside `RegistrationBuilder` to `protected internal`. This should allow classes that inherit from `RegistrationBuilder` to have access to those fields. 

I also changed the visibility of `InjectorCache` to `public`. This is needed for the cases that a custom `RegistrationBuilder` needs to be made.


hadashiA:
I too have been having problems with extensibility lately. Thanks.
I'll fix the namespace later.

PR #565, opened 2023-10-30
Allow installers to be added to localExtraInstallers

pnarimani:
As it is, even though `localExtraInstallers` is a list, it's not possible to add more than one installer to it. 

I added a `EnqueueLocalInstaller` function to `LifetimeScope`, which allows users to add more installers to `localExtraInstallers` list. This function returns an `IDisposable` struct which removes the added installer.

### DIFF
--- a/VContainer/Assets/VContainer/Runtime/Unity/LifetimeScope.cs
+++ b/VContainer/Assets/VContainer/Runtime/Unity/LifetimeScope.cs
@@ -47,6 +47,23 @@ namespace VContainer.Unity
             }
         }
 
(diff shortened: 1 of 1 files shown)


AlonTalmi:
From what I see the localExtraInstallers is pretty much a hack used for testing I think it would better to deprecate it then extend it :(

What do you use it for? 
When creating new LifetimeScopes, the recommended way is by Enqueuing to the GlobalExtraInstallers, which does support multiple installers

hadashiA:
Thank you. But I agree with AlonTalmi's view.
I am not too positive about breaking Installer into smaller pieces.

Issue #387, opened 2022-05-19
VContainer Diagnostics window


Yggdrasills:
Hey! The problem is that references to already removed dependencies do not disappear in the editor window. For example, you can create a `LifeTimeScope` with one registration and one entry point, which in the constructor accepts the created registration. Run the scene -> remove the `LifeTimeScope` object from the scene (it will call `Dispose` on `OnDestroy`) -> look at the diagnostic window -> dependencies are also present, where it will also be seen that there are links to the created dependencies.

It would be much more convenient if the diagnostic window displayed real-life dependencies at runtime

```
public class DummyScope : LifetimeScope
    {
        protected override void Configure(IContainerBuilder builder)
        {
            builder.Register<DummyModel>(Lifetime.Singleton);
            builder.RegisterEntryPoint<DummyController>();
        }
    }

public class DummyModel
    {
    }

public class DummyController : IInitializable
    {
        private DummyModel _model;

        public DummyController(DummyModel model)
        {
            _model = model;
        }

        public void Initialize()
        {
        }
    }
```

Please add a Refresh button next to the Reload button which will list only existing dependencies

hadashiA:
Perhaps it was solved in #568.
If there is a problem, please reopen.

PR #568, opened 2023-11-10
Fix diagnostic windows are not updating realtime

PeraSite:
1. When a container is disposed, the diagnostics window still shows the lifescope and its dependencies. (#387)
In this situation, when user clicks scope/dependency that already disposed to view the details, it shows an error.
이미지

This PR fixes the error by clearing the diagnostics information when a container is disposed.

2. On editor's play mode start / end, diagnostic windows are not automatically updated.
Using `EditorApplication.playModeStateChanged`, diagnostic windows are conveniently updated.


hadashiA:
👍

PR #569, opened 2023-11-14
Added RegistrationBuilder variations with callback support

AlonTalmi:
I added the option to listen to resolve callback when binding new types.
It's useful if you want to use the instance when it's created without coupling it to the type implementation itself.

I decided to implement it using new variations of RegistrationBuilder, one is called RegistrationBuilderWithCallback the other is called FuncRegistrationBuilderWithCallback, no need for instance callback.

The con of doing it this way is that there's more permutations if someone want to create new type of RegistrationBuilders and still support callbacks.
 
The other ways I could have implemented it is:
- Adding a WithCallback method in the base RegistrationBuilder, but because I don't know the type of the object I could only use Action<object, IObjectResolver>
- By creating a decorator to the final created Registration by extracting an interface, but then I would have had to replace every reference to Registration to IRegistration (which also breaks backwards compatibility)

Tested on my game and wrote Unit Tests

### DIFF
--- a/VContainer/Assets/VContainer/Tests/ContainerTest.cs
+++ b/VContainer/Assets/VContainer/Tests/ContainerTest.cs
@@ -518,5 +518,35 @@ namespace VContainer.Tests
             var ctorInjectable = new ServiceA(new NoDependencyServiceA());
             Assert.DoesNotThrow(() => container.Inject(ctorInjectable));
         }
+
+        [Test]
+        public void RegisterWithCallback()
+        {
+            {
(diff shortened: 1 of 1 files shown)


hadashiA:
I think it's good, but if the `callback` argument of the API is just added, it's not obvious when it will be called.
Currently, lambda expressions tend to be used primarily for additional configuration to be done programmatically and not for callback purposes.

`Register(..). .OnResolve(callback)` form would be better, or something like that. I wonder.

PR #570, opened 2023-11-14
Added option to lazily create paramaters and use the scope's resolver

AlonTalmi:
Changed the interface IInjectParameter ( from: Value { get; }  to: GetValue(IObjectResolver resolver) )

And added new variations of it called FuncTypedParameter and FuncNamedParameter.

This allows you to add parameters to registrations the can be created lazily/created per transient resolve and can use the IObjectResolver that was used to create the object.

It would be nice to also add in the future the the option to declare a parameter as single (in case the parent register is not)

Added Unit Tests

Edit: Also added option to resolve a parameter from an injector in a separate [branch](https://github.com/AlonTalmi/VContainer/tree/InjectedInstanceParameter)

### DIFF
--- a/VContainer/Assets/VContainer/Runtime/IInjectParameter.cs
+++ b/VContainer/Assets/VContainer/Runtime/IInjectParameter.cs
@@ -5,6 +5,6 @@ namespace VContainer
     public interface IInjectParameter
     {
         bool Match(Type parameterType, string parameterName);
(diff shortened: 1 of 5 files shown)


hadashiA:
Thanks a lot.
I am really sorry for the delay in checking.

I will try to merge this.

hadashiA:
#640

PR #571, opened 2023-11-14
IContainerBuilder RegisterOnDisposeCallback option

AlonTalmi:
I added the option to listen to when any container created by the builder is disposed (but not scoped child containers).

I added a RegisterDisposeCallback to interface IContainerBuilder.

The dispose callback is called just before the container is disposed.

Added Unit Tests

### DIFF
--- a/VContainer/Assets/VContainer/Runtime/Container.cs
+++ b/VContainer/Assets/VContainer/Runtime/Container.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 using VContainer.Diagnostics;
 using VContainer.Internal;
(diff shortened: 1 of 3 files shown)


hadashiA:
@AlonTalmi The need is a bit subtle to me, but since there is no easy way to do this, I think it is a good idea. Thanks.

If you'd like, I'd be happy to resolve the conflicts on this as well.

Sorry for the delay in response!

VContainer/Assets/VContainer/Runtime/Container.cs
```diff
@@ -55,17 +56,20 @@ public sealed class ScopedContainer : IScopedObjectResolver
...
         readonly Func<Registration, Lazy<object>> createInstance;
+        readonly IEnumerable<Action<IObjectResolver>> disposeCallbacks;
 
         internal ScopedContainer(
             Registry registry,
             IObjectResolver root,
             IScopedObjectResolver parent = null,
-            object applicationOrigin = null)
+            object applicationOrigin = null,
+            IEnumerable<Action<IObjectResolver>> disposeCallbacks = null)
```
hadashiA, line 66: I also noticed recently that if you only keep the Action<IObjectResolver> disposeCallback type perhaps,

```cs
disposeCallback += callback
```

I think it should be possible to register more than one callback, as in
You can verify this if you like.

VContainer/Assets/VContainer/Runtime/Container.cs
```diff
@@ -110,6 +114,8 @@ public bool TryGetRegistration(Type type, out Registration registration)
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public void Dispose()
         {
+            EmitDisposeCallbacks();
```
hadashiA, line 117: Is there any way to add values to disposables instead of holding dispoeCallbacks?

For example.

```cs
struct ActionDisposable : IDisposable
{
    Action calback;

    public ActionDisposable(Action callback)
    {
        this.callback = callback; }
    }

   public void Dispose() => callback();
}

disposables.Add(new ActionDisposable(callback));
````


Translated with www.DeepL.com/Translator (free version)

hadashiA:
#603
I have created a PR on my end that incorporates your suggestions. Thanks a lot.

PR #572, opened 2023-11-14
Support for Value Types delegate registration

AlonTalmi:
Re-implementation of #521 

Also added Unit Tests


hadashiA:
merged in #641

PR #573, opened 2023-11-15
Added option to TryGetSharedInstance to IObjectResolver

AlonTalmi:
Added to IObjectResolver interface a TryGetSharedInstance which only returns if a shared instance of the request type (or Registration) exists and was created.

Can be used to update instances of a type (if they're not transient) at runtime without resolving them needlessly.

I also added generic extension methods and Unit Tests


hadashiA:
The use case may not be very clear. Also, I am reluctant to function in a way that could abuse DI. Please let me reject this time. Thanks for your suggestion.

PR #574, opened 2023-11-15
Added AsNoneLazy ContainerBuilder extension

AlonTalmi:
I added a simple extension method that registers a build callback that resolves the type requested.

Unlike Zenject and other similar DI libraries where NoneLazy is marked during the registration phase here it's controlled in the container level. I personally like it, but can see why you might not.

Anyway, I also added a Unit Test.

### DIFF
--- a/VContainer/Assets/VContainer/Runtime/ContainerBuilderExtensions.cs
+++ b/VContainer/Assets/VContainer/Runtime/ContainerBuilderExtensions.cs
@@ -130,6 +130,13 @@ namespace VContainer
             Func<IObjectResolver, Func<TParam1, TParam2, TParam3, TParam4, T>> factoryFactory,
             Lifetime lifetime)
             => builder.Register(new FuncRegistrationBuilder(factoryFactory, typeof(Func<TParam1, TParam2, TParam3, TParam4, T>), lifetime));
+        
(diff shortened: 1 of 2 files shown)


hadashiA:
Thank you. Probably useful.

However, I don't think the name NonLazy is very clear.

If it is not resolved from anywhere and an Instance is created, it is nothing more than an entry point in the broadest sense.
So I have only allowed APIs that specifically specify the type of entry point, such as `RegisterEntryPoint` or `RegisterComponent`.

Maybe it's the same with ASP .
NET Core, for example AddControllers, AddHostedService, etc.
I prefer that approach.

Ceeeeed:
@hadashiA 
Sorry for reviving an old closed PR, but this is still a relevant topic for my project.

I think this feature would be great.
Right now, I’m using `RegisterEntryPoint()` with an empty `Initialize()` function from the `IInitializable` interface just so the object gets created after the container is built (I’m subscribing to events in that object’s constructor, so I want it to be instantiated immediately).

With `NonLazy()` (or whatever you’d prefer to call it if you don’t like that name), I could avoid creating an empty function just for that.

If I’m approaching this the wrong way and there’s a better pattern for this use case, please let me know.

Issue #540, opened 2023-07-31
[Inject] called after OnEnable


YegorStepanov:
The following code works in Zenject, but not in VContainer:
```C#
[Inject]
private void Construct(IRepository repository)
    _repository = repository;

private void OnEnable() =>
    _repository.Add(item); // VContainer: NRE
        
private void OnDisable() =>
    _repository.Remove(item);
```

Execution order of VContainer:
```
Awake
OnEnable
Construct
```

Execution order of Zenject:
```
Awake
Construct
OnEnable
```

VContainer uses the call order according to the unity documentation, but it is not convenient. It also makes it difficult to switch from Zenject.

To get around this, Zenject uses a hack:
https://github.com/modesttree/Zenject/blob/31f06cf81043f04301b1072ec81922b64149ea34/UnityProject/Assets/Plugins/Zenject/Source/Main/DiContainer.cs#L1680

AlonTalmi:
I made my own version of this hack as extension method:
```c#
public static GameObject InstantiateAndInject([NotNull] this IObjectResolver resolver, [NotNull] GameObject prefab, Transform parent = null)
        {
            if (prefab == null)
                throw new NullReferenceException(nameof(prefab));
            
            var prefabActive = prefab.activeSelf;
            
            prefab.SetActive(false);
            var instance = resolver.Instantiate(prefab, parent);
            prefab.SetActive(prefabActive);
            
            instance.SetActive(prefabActive);
            return instance;
        }
```

YegorStepanov:
Thank you very much for the code!

It would be great if VСontainer did it itself.

alireza-entertale:
Hi,
I didn't have any issue about calling order of `OnEnable` and `[Inject] Construct`. `OnEnable` was called before `[Inject] Construct` but after using RootLifetimeScope, `OnEnable` is called before `[Inject] Construct`. Is it the same issue here ?
I already fixed that by moving the code from `OnEnable` to `Start` and it is okay in my case but I'm curious why it happens after trying to use RootLifetimeScope.

PR #575, opened 2023-11-15
Inject before awake when instantiating

AlonTalmi:
Solves #540

The object is instantiated disabled and only enabled after it was injected, that way awake is only called after injection.

Added to the existing Unit Test an assert proving the bug


hadashiA:
@AlonTalmi  👍 
Thanks for the fixing. I think it is a good improvement.
I was going to merge, can I get the conflicts resolved?

hadashiA:
I was able to follow your commit and resolve the conflict. I've merged it here. Thanks.

PR #576, opened 2023-11-15
Exposed IObjectResolver to prefab and parent finders in ComponentRegistrationBuilder

AlonTalmi:
When registering a component the prefab and parent had to be known, I added the option to declare them using a delegate with the IObjectResolver

Also have a [branch](https://github.com/AlonTalmi/VContainer/tree/SeparateComponentRegistrationBuilder) that separates the ComponentRegistrationBuilder into separate classes


hadashiA:
I will adopt it. Thank you.

PR #578, opened 2023-11-16
 Fix withParameter not working (With code generation)

AlonTalmi:
Includes @adarapata's commit in PR #457, updated to master and applied to code generation as well.

I had some problems testing it, but from what I understood it seemed to be working


hadashiA:
Ok, thanks for the fix 👍

PR #579, opened 2023-11-16
Installer component (like Zenject's MonoInstaller)

AlonTalmi:
This is the equivalent of Zenject's MonoInstaller

Also added:
InstallerGroup component that batch multiple installers together (I mostly used it in prefabs, that way I can create a group at the root of the prefab and I don't need to reference the nested installers in the prefab 1 by 1)

GetInstallerComponentsGroup component that batches all installers in GameObject using GetComponent

BuildCallbackInstaller component that is used for actions that run on the Container's build callback

### DIFF
--- a/VContainer/Assets/VContainer/Runtime/Unity/InstallerComponents.meta
+++ b/VContainer/Assets/VContainer/Runtime/Unity/InstallerComponents.meta
@@ -1,0 +1,3 @@
+fileFormatVersion: 2
+guid: 654eab9df1ff4342aa6dc8b27e7daa6b
+timeCreated: 1700142729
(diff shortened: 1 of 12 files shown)


pnarimani:
I don't think this is necessary to be honest. You can always create a LifetimeScope that does a GetComponent or whatever it needs to do in the `Configure` method.

AlonTalmi:
Maybe the name is confusing, this is not an installer that specifically installs components, but an abstract MonoBehaviour component that inherits from IInstaller.

I like to split my LifetimeScopes into multiple installers, which many times can be recycled in other scopes and reduce boilerplate.

srodriguezf:
Hi, we are thinking changing to VContainer. Now, we are using Zenject. This feature is very important for us because we use a lot of installers and we cant write everything in the lifetimeScope. Would like when it is going to be the merge pull request. Thanks.

hadashiA:
Thanks.

Several people seem to be in favour of this feature,
Regrettably, I do not actively recommend separating DI settings.
So, I would like to reject this. Sorry.
(The `OnBuild` method, however, might not be a bad idea.)

As discussed in this thread, if users want to separate them in any way they like, they can.

PR #580, opened 2023-11-19
System.Object always passed to InjectTypeInfo

AlonTalmi:
Small bug fix where System.Object always passed to InjectTypeInfo.

This only affected error messages, failed to resolve errors always shown: "Failed to resolve System.Object" instead of the actual type, really annoying when debugging.


hadashiA:
Thanks 👍

PR #585, opened 2023-11-30
Added a TryResolve method to IObjectResolver

AlonTalmi:
I added to IObjectResolver the method:
bool TryResolve(Type type, out object resolved);

And also added 2 extension methods:
TryResolve<T>
T TryResolveOrDefault<T>(T defaultValue = default)

I also implemented it in EntryPointDispatcher because it used try-catch to mimic TryResolve

And added few simple unit test (copied from the default resolve method unit tests)

Also allows to add an Optional inject attribute


hadashiA:
👍

hadashiA:
Merged in #656

PR #586, opened 2023-12-03
Fixed LifetimeScope's Container is null during build callback

AlonTalmi:
If you try and use the LifetimeScope's Container property during the container's build callback (for example when using the Instantiate extension method), the Container is null.

That happens because the Container property is assigned only after the Build method returns (which invokes the build callbacks).

As a quick fix I registered a build callback which registered the container to the Container property (before installing all other installers), that works because the build callback are called FIFO.

Edit: now also fixes AutoInjectGameobjects called after build callbacks which can easily cause duplicate injections if you instantiate an injected GameObject in a build callback


hadashiA (approved):
👍 Thanks

PR #629, opened 2024-02-14
Add using new faster unity find object method

LovorDev:
Under the preprocessor, a new unity  method has been added that searches for the first available object. This new function runs 2x faster than the old one. It has been implemented starting from these versions of Unity: 
2023.1.0a20, 
2022.2.5f1, 
2021.3.18f1,
2020.3.45f1.
Defines UNITY_2020_4_OR_NEWER || UNITY_2021_4_OR_NEWER are not working at the moment, but they have been added for future use.


hadashiA:
Makes sense, thanks!

PR #638, opened 2024-02-29
Resolve time diagnostics

AlonTalmi:
Transient Uses average time
Singleton and scoped use peak time


hadashiA:
👍

AlonTalmi:
@hadashiA Maybe I'm a bit late, but I feel the fact that the time's behavior is different per lifetime is unclear.
Maybe it would be better to have 2 columns, one with average resolve time and one with peak resolve time?

PR #640, opened 2024-03-02
With parameter resolver

hadashiA:
port from #570


christopher-qde:
Just a FYI, this ended up being a breaking change in my current project.

`error CS0121: The call is ambiguous between the following methods or properties: 'RegistrationBuilder.WithParameter<TParam>(TParam)' and 'RegistrationBuilder.WithParameter<TParam>(Func<IObjectResolver, TParam>)'`

Issue #621, opened 2024-01-25
VContainer Diagnostics shows partial information


rasentry:
First of all, thank you for this amazing project.

The problem is: When several scenes are loaded and both of them have different LifetimeScopes attached to GameObjects with the same name, VContainer Diagnostics shows information only about one of them.

hadashiA:
Probably solved in #658.

PR #655, opened 2024-03-29
Fix Diagnostic window.

DenisPimenov:
Hi @hadashiA ! 
I have the problem described here https://github.com/hadashiA/VContainer/issues/621. I've been looking at the logic for collecting diagnostics and was going to suggest the simple solution, to check the unique scopes with GameObject instanceId.
There is a small problem from using Lookup when building a window. For this I had to add an instanceId to the name of the scope to output to the window.

Also wanted to suggest to add cleaning the scopes from the diagnostics after Dispose/Destroy.

If you have any other suggestions I can fix this solution.

### DIFF
--- a/VContainer/Assets/VContainer/Runtime/Diagnostics/DiagnosticsContext.cs
+++ b/VContainer/Assets/VContainer/Runtime/Diagnostics/DiagnosticsContext.cs
@@ -6,19 +6,19 @@ namespace VContainer.Diagnostics
 {
     public static class DiagnositcsContext
     {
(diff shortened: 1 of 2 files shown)


hadashiA:
Thank you!

However, there was a slight point of concern.
 I want the DiagnosticsContext to be able to collect information in a Unity-independent layer. So I don't want to rely too much on the concept of instanceId.
After all, I don't think I need to increase the argument for GetCollector, since the string will still be the key.

I will probably rework that after your commit.
Thanks a lot.

PR #660, opened 2024-04-23
Added test and fix for when a game object throws during instantiation

NordlightPer:
In our game we sometimes have problems during the init of a game mode. When this happens, we have a transition game object that is disabled and occasionally that change gets committed. Similar try finally guards other code paths but not the one we were using.
This fixes our case but maybe all paths should have this construct?


hadashiA:
This change is probably desirable. Thanks a lot.

PR #666, opened 2024-05-13
Fix CappedArrayPool some bug

Ruoh3kou:
Fix CappedArrayPool :
1. alloc new array without caching 
3. some cached array is not be used


hadashiA:
Oh. Looks like you've made your point.
Thank you very much.

PR #679, opened 2024-06-06
Make IAsyncStartable available on Unity 2021.3 using System.Threading.Tasks.Task

left-:
This PR changes IAsyncStartable to support UniTask, .NET Task, and Awaitable.

It makes async startable available to projects running on current Unity LTS that cannot use UniTask (e.g., due to legacy reasons).

It also adds a new entry point test to IAsyncStartable and fixes a bug in which tasks throw exceptions before executing a continuation.


hadashiA:
Looks good. Thanks !

PR #681, opened 2024-06-16
Fix for usage of ListBuffer to be properly released

Raegan03:
When checking the VContainer implementation I've found that ListBuffer is used incorrectly in two instances. The list is being provided by the pool but it's never released to it. I've made a quick fix using BufferScope to properly dispose of and release the list back to the pool.

I've done a test with `LifetimeScope:Find<T>(Scene scene)` and for 1000 calls only one list was created as expected.


hadashiA:
Thanks!

Issue #686, opened 2024-06-24
[Bug] EntryPoint is called twice


kochounoyume:
This phenomenon occurred in the latest version 1.16.0.

How to reproduce:

1. Lay the scene over the hierarchy in this way before the game is played.
![image](https://github.com/hadashiA/VContainer/assets/78918084/2627d429-86a3-4c09-8585-c205de904cb1)

2. Place a LifetimeScope in each scene and associate it with a parent-child relationship setting.
![image](https://github.com/hadashiA/VContainer/assets/78918084/e9dd69cd-e55e-453f-bda5-9867c1d05dcf)
![image](https://github.com/hadashiA/VContainer/assets/78918084/ddd6673a-24a5-483c-9296-55b607c0a002)
![image](https://github.com/hadashiA/VContainer/assets/78918084/ae7100c6-69e9-49e4-903e-4840ca5be71a)

3. Register EntryPoints with ChildLifetimeScope.
```cs
using VContainer;
using VContainer.Unity;

public sealed class ChildLifetimeScope : LifetimeScope
{
    protected override void Configure(IContainerBuilder builder)
    {
        builder.RegisterEntryPoint<EntryPointTest>();
    }
}
```

```cs
using UnityEngine;
using VContainer.Unity;

public sealed class EntryPointTest : IStartable
{
    void IStartable.Start()
    {
        Debug.Log("Startable!");
    }
}
```

4. Play the game.
![image](https://github.com/hadashiA/VContainer/assets/78918084/7634969a-9981-408f-a6ea-bd44fcd49d1f)

PR #687, opened 2024-06-24
Fix unintended multiple registrations

Akeit0:
Fixed multiple registrations which occurs when the scope has grandparent scope.
https://github.com/hadashiA/VContainer/issues/686


Raegan03:
I've noticed it today as well, thanks for the fix! 💪

hadashiA:
Oh. That's my mistake ><  Thanks a lot!

PR #688, opened 2024-06-29
fix source generator for public property private setter

TS696:
Fixed an issue where the source generator would generate code that accesses inaccessible property setters.


hadashiA:
👍

Issue #669, opened 2024-05-24
If a subclass defines a private field with the same name as one in the baseclass, it will cause some private fields in the baseclass not to be injected.


1000111001:
Test code:
```CSharp
public class BaseClass
{
    [Inject] private readonly int _privateReadonlyFieldValue;
    [Inject] private int _privateFieldValue;
    [Inject] private int _privateFieldValue2;


    public void AreEqual(int actual)
    {
        Assert.AreEqual(_privateReadonlyFieldValue, actual);
        Assert.AreEqual(_privateFieldValue, actual);
        Assert.AreEqual(_privateFieldValue2, actual);
    }
}

public class SubClass : BaseClass
{
    [Inject] private int _privateFieldValue;
}
```
`_privateFieldValue` and `_privateFieldValue2` in `BaseClass` will not be injected.

hadashiA:
#691, this case now throws an exception.

Declaring fields with exactly the same signature in SubClass is nonsense.

PR #691, opened 2024-07-02
Fix duplicate injection issue + introduce Editor tests

in3orn:
Addressing issue #669:
Currently if `TypeAnalyzer` encounters duplicate field (with the same `Name`) in a base class, it skips injection for this class. As a result all the following fields of a base class are not injected. It leads to unexpected runtime behaviour. There are two solutions:
- supporting "duplicate injections" like Zenject does
- throwing exception to give immediate feedback that Injection process will go wrong

I did the 2nd approach as it detects developer's mistakes that in most cases can be easily fixed by protected injections.

Ps. I've also introduced the Editor tests. Is there any reason why most tests have to be run in a Playmode? If you like I can tweak & merge my test with others related to the TypeAnalyzer. Just wanted to know the approach :)


hadashiA:
Thanks for the PR 🙏 

> Ps. I've also introduced the Editor tests. Is there any reason why most tests have to be run in a Playmode? If you like I can tweak & merge my test with others related to the TypeAnalyzer. Just wanted to know the approach :)

Indeed, most may not need to be in Play mode.
However, there are tests in Play mode, so I preferred to have one test assembly rather than multiple separate ones.

Maybe we could split them up as you suggest.
However, in this PR, the `.UnitTests` asmdef has been added, but in fact the current `.Tests` is also a Unit Test, so maybe the name is a bit inappropriate?
If there are no strong opinions I think we could leave it as just Play mode, but what do you think?

in3orn:
I think we could rename `VContainer.Tests` to `VContainer.PlayModeTests` and `VContainer.UnitTests` to `VContainer.EditModeTests`. However moving `DuplicateInjectionTest` to `VContainer.Tests` is also fine for me. I can also move it to `TypeAnalyzerTest` if you think my fix is ok and don't want to rework the flow in that regard.

hadashiA:
>  I can also move it to TypeAnalyzerTest if you think my fix is ok and don't want to rework the flow in that regard.

I would like to treat the renaming of asmdef as a separate matter, so I would appreciate it if this PR did this.

in3orn:
Done :)

hadashiA:
Ok, thx!

PR #692, opened 2024-07-02
feature: Add protected FindParent method

gtg092x:
This will allow implementors the opportunity to override runtime Find behavior. This is particularly useful for projects that implement multiple potential parent scopes and need a way to disambiguate them.

For example, some child scopes might want to specifically find ancestors up their transform tree:

```csharp
public class ParentedLifetimeScope : LifetimeScope {
    protected override LifetimeScope FindParent() {
        LifetimeScope[] allScopes = FindObjectsByType(type, FindObjectsSortMode.None).Cast<LifetimeScope>().ToArray();
        return allScopes.First(x => IsAncestor(x.gameObject));
    }

    public bool IsAncestor(GameObject target) {
            var parent = this.transform.parent;
            while (parent != null)
            {
                if (parent == target.transform)
                {
                    return true;
                }

                parent = parent.parent;
            }

            return false;
        } 
}
```

All unit tests pass with this change and because the implementation is a virtual method that defaults to the previous functionality, current behavior is unchanged.

### DIFF
--- a/VContainer/Assets/VContainer/Runtime/Unity/LifetimeScope.cs
+++ b/VContainer/Assets/VContainer/Runtime/Unity/LifetimeScope.cs
@@ -307,12 +307,24 @@ namespace VContainer.Unity
             EntryPointsBuilder.EnsureDispatcherRegistered(builder);
         }
 
+        protected virtual LifetimeScope FindParent() => null;
+
         LifetimeScope GetRuntimeParent()
         {
             if (IsRoot) return null;
 
             if (parentReference.Object != null)
                 return parentReference.Object;
+            
+            // Find via implementation
+            var implParent = FindParent();
+            if (implParent != null)
+            {
+                if (parentReference.Type != null && parentReference.Type != implParent.GetType()) {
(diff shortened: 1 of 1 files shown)


hadashiA:
It is a good idea.

However, with this implementation, it seems that the following conditions have to be met.

```cs
if (parentReference.Type != null && parentReference.Type != GetType())
```

Shouldn't it be possible to override with or without parentReference.Type?

gtg092x:
That's true. I'll make a small design change so that the override can be either.

hadashiA:
Thank you very much. I think it's good.

The only point, is that it doesn't seem very clear from the point of view of the user considering override how `FindParent()` and `FindParent(Type)` should be used.

Since `FindParent()` probably also serves as a `FindParnet(Type)`, why not just have the former be virtual?

gtg092x:
Thank @hadashiA! I had to fight my commit history a little bit to get a clean PR, but this should simplify the interface.

hadashiA:
👍  Thanks for the fixing.

PR #706, opened 2024-09-10
Update register-monobehaviour.mdx

dmtr-t:
Method names fixes


hadashiA:
Thanks!

hadashiA:
I got the impression that the name of this document might be better...

PR #712, opened 2024-10-10
Allow scope naming on create

Geminior:
A small addition to allow custom names to LifetimeScopes when manually created through calls to Create(...) or CreateChild(...)

Also makes the Create(IInstaller) overload public as that seems to be an oversight?


hadashiA:
Makes sense, thanks 👍

PR #741, opened 2025-01-06
Adding Singleton Lifetime Scope

IShix-g:
Although VContainer has successfully been kept simple, I believe adding this extension could be beneficial. Therefore, I am submitting this request for your consideration. Thank you for reviewing it.

## Reasons for the Addition

This extension proves useful in cases where the [Project root LifetimeScope](https://vcontainer.hadashikick.jp/en/scoping/project-root-lifetimescope) cannot be utilized for some reason.

- Enables initialization in an arbitrary scene, instead of during game startup.
- Since it needs to be placed in the hierarchy, it provides better visibility.

## Advantages

- Offers a `LifetimeScope` that persists throughout the game without being destroyed.
- Useful for writing processes that should only be initialized once, such as ad initialization or save data setup.
- Since one-time initialization processes can be handled, there’s no need for repeated initialization in subsequent scenes, improving performance.
- By placing it in every scene, it ensures proper functioning even when starting the game from any scene. If it is not placed in all scenes, the game can only be started from the title scene (the initial scene).

### Two Classes Implemented

- `SingletonLifetimeScope<T>`: Acts as the parent singleton `LifetimeScope`.
- `NestedSingletonLifetimeScope<T>`: Acts as a child `LifetimeScope`.

### SingletonLifetimeScope

Due to the singleton nature of this class, the `Parent` option is always set to `None` and cannot be modified.

Src4

### NestedSingletonLifetimeScope

The `SingletonLifetimeScope` works independently, but if you need to use child `LifetimeScopes`, you can implement the `NestedSingletonLifetimeScope`. This class allows specifying the parent type during implementation, making it error-proof. It is particularly useful when implementing `LifetimeScopes` for each scene.

The `Parent` option can only be set to the designated parent type and cannot be modified otherwise.

Src3

## Testing

All generation tests passed successfully.

You can review the tests here:
[Assets/Tests/Unity/SingletonLifetimeScopeTest.cs](https://github.com/hadashiA/VContainer/pull/741/commits/e541668ae2a41f45bfe0fcb6545c7d85aa237b7b)

## Fixes to the ParentReferencePropertyDrawer

Both `SingletonLifetimeScope` and `NestedSingletonLifetimeScope` are `abstract classes`. However, they were still being displayed in the inspector. Since selecting them in the inspector would lead to errors, a [fix has been implemented to hide them](https://github.com/hadashiA/VContainer/commit/669f96d8134b4ad0d770f5122e30edd96f697487).

Src1

### DIFF
--- a/VContainer/Assets/Tests/Unity/Fixtures/SampleNestedSingletonLifetimeScope.cs
+++ b/VContainer/Assets/Tests/Unity/Fixtures/SampleNestedSingletonLifetimeScope.cs
@@ -1,0 +1,13 @@
+using VContainer.Unity.Extensions;
+
+namespace VContainer.Tests.Unity
+{
+    public sealed class SampleNestedSingletonLifetimeScope : NestedSingletonLifetimeScope<SampleSingletonLifetimeScope>
+    {
+        protected override void Configure(IContainerBuilder builder)
+        {
+            builder.Register<DisposableServiceA>(Lifetime.Scoped);
+            builder.Register<DisposableServiceB>(Lifetime.Singleton);
+        }
+    }
+}
--- a/VContainer/Assets/Tests/Unity/Fixtures/SampleNestedSingletonLifetimeScope.cs.meta
+++ b/VContainer/Assets/Tests/Unity/Fixtures/SampleNestedSingletonLifetimeScope.cs.meta
@@ -1,0 +1,3 @@
+fileFormatVersion: 2
+guid: 9ac8ceb41a60446cb34ff54fa5a6bc20
+timeCreated: 1736124224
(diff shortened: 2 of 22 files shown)


IShix-g:
I am creating a custom editor extension to make the `LifetimeScope`'s Parent property intentionally read-only.
It seems that specifying a Generic Type in `CustomEditor` no longer works in Unity 2022.
I will investigate further and make the necessary fixes.

[feat: Modify singleton lifetime scopes to disable the selection of ParentReference.](https://github.com/hadashiA/VContainer/pull/741/commits/89004ea583ad830733965a466b5569a33df89c20)

PR #744, opened 2025-01-12
Apply VContainerSettings' RemoveClonePostfix to the root lifetime scope

jonah1nine:
Made it so the `VContainerSettings`' `RemoveClonePostfix` setting is also applied to the root lifetime scope.


hadashiA:
👍

Issue #739, opened 2024-12-28
Creating a child scope from root scope does not parent the child scope game object to the root scope game object


jonah1nine:
When creating a childscope using 
```
rootScope.CreateChild();
```
the child scope game object is not parented to the root scope game object. 
However creating a child scope from prefab, parents the child scope game object to the root scope game object as expected.
```
rootScope.CreateChildFromPrefab(childScopePrefab);
```
---
Also if a non-root scope creates a child scope, it parents the child scope game object to parent scope game object as expected.
```
nonRootScope.CreateChild();
```
---
This is what I get when running the code below
<img width="284" alt="Screenshot 2024-12-28 at 12 21 44" src="https://github.com/user-attachments/assets/c87111bf-90c0-4278-a4e5-ba593a1ad4ab" />
```CSharp
public class Test : MonoBehaviour
{
    [SerializeField] private ChildLifetimeScope childLifetimeScopePrefab;
    
    void Start()
    {
        var rootScope = LifetimeScope.Find<RootLifetimeScope>();
        var parentScope = LifetimeScope.Find<ParentLifetimeScope>();

        rootScope.CreateChild(childScopeName: "Child Lifetime Scope - Root");
        rootScope.CreateChildFromPrefab(childLifetimeScopePrefab);
        parentScope.CreateChild(childScopeName: "Child Lifetime Scope - Parent");
    }
}


public class RootLifetimeScope : LifetimeScope { }
public class ParentLifetimeScope : LifetimeScope { }
```
---
Also is setting the parent of a LifetimeScope in editor to the root scope makes that LifetimeScope game object parented to the root scope? Because that's not happening as well - as seen above.
<img width="307" alt="Screenshot 2024-12-28 at 12 33 27" src="https://github.com/user-attachments/assets/4bca104b-4259-4a8f-98f5-f10c52c22c31" />

PR #760, opened 2025-03-22
Issue#739 Always parents created child lifetime scope game object

jonah1nine:
PR to addres this issue: https://github.com/hadashiA/VContainer/issues/739

If the parent lifetime scope is the root, parenting the child lifetime scope game object to it would automatically make it "not be destroyed on load" (DontDestroyOnLoad).

This change results in all `CreateChild` api behave the same with regards to their game object parent/child relationship (previous discrepancy was outlined in the issue linked above).


hadashiA:
ok, thank you very much. Sorry for the late response.

PR #788, opened 2025-07-06
Fixed compilation error on older unity versions

migus88:
After the merge of Keyed registrations PR, I noticed that some checks haven't passed. 
Looked into it and it's due to older C# version in earlier Unity versions. 
This minor fix should do it.


hadashiA:
thanks 👍

Issue #157, opened 2021-03-10
Decorator support?


Razenpok:
Scrutor library adds a cool decorator feature for `Microsoft.Extensions.DependencyInjection`. Is it possible to provide a similar thing here?

https://github.com/khellang/Scrutor

```csharp
builder.Register<IPictureRepository, PictureRepository>();
builder.Decorate<IPictureRepository, CachingPictureRepository>();
builder.Decorate<IPictureRepository, LoggingPictureRepository>();

// Resolve<IPictureRepository> would resolve LoggingPictureRepository with this hierarchy of decoration
// - LoggingPictureRepository
// | - CachingPictureRepository
//   | - PictureRepository
```

hadashiA:
Ok, this may be useful.
I think the Decorater pattern needs some support on the DI library side.
( Because we need to resolve the same interface).

This may require a little work on the core functionality, so let me consider it.

However, I may prefer to make this like a plugin.

vu-truong:
Any update?

hadashiA:
In #372, I am trying to add an interface that will allow the ConotainerBuilder to be modified.
I believe this will allow for the implementation of perhaps the following:

```csharp
        public void RegisterDecorator(Type interfaceType, Type decoratorType)
        {
            for (var i = 0; i < Count; i++)
            {
                var entry = this[i];
                var isTarget = entry.ImplementationType.IsInterface &&
                               entry.ImplementationType == interfaceType;

                if (entry.InterfaceTypes is { } interfaceTypes)
                {
                    foreach (var t in interfaceTypes)
                    {
                        if (t == interfaceType)
                        {
                            isTarget = true;
                            break;
                        }
                    }
                }

                if (isTarget)
                {
                    this[i] = new DecoratorRegistrationBuilder(this[i], decoratorType); 
                }
            }
        }
```

```csharp
    public class DecoratorRegistrationBuilder : RegistrationBuilder
    {
        readonly RegistrationBuilder inner;
        readonly Type decorateType;
        
        public DecoratorRegistrationBuilder(RegistrationBuilder inner, Type decoratorType) 
            : base(decoratorType, inner.Lifetime)
        {
            this.inner = inner;
            decorateType = inner.InterfaceTypes != null ? inner.InterfaceTypes[0] : inner.ImplementationType;
            InterfaceTypes = inner.InterfaceTypes;
            As(decorateType);
        }
    
        public override Registration Build()
        {
[...]

PR #790, opened 2025-07-17
Fix decorator

CncealDVRflectN:
- Fix for decorator registration (by implementation class as well as by factory function)
- Add support for nested decorators as was highlighted in the initial proposal in #157 
- Add test cases

Should I test some other cases? Also, if there are no changes needed to push it to production, I can merge changes from `master` and retarget this PR to it

### DIFF
--- a/VContainer/Assets/Tests/DecoratorTest.cs
+++ b/VContainer/Assets/Tests/DecoratorTest.cs
@@ -50,12 +50,46 @@ namespace VContainer.Tests
         }
 
         [Test]
+        public void Decorate_Func()
+        {
+            var builder = new ContainerBuilder();
+            builder.Register<IDecoratedType, DecoratedType>(Lifetime.Singleton);
+            builder.RegisterDecorator<IDecoratedType, Decorator1>((original, container) => new Decorator1(original));
+
+            var container = builder.Build();
+            var instance = container.Resolve<IDecoratedType>();
+            Assert.That(instance, Is.TypeOf<Decorator1>());
+            
+            var inner = ((Decorator1)instance).Inner;
+            Assert.That(inner, Is.TypeOf<DecoratedType>());
+        }
+
+        [Test]
         public void Decorate_Nested()
         {
             var builder = new ContainerBuilder();
             builder.Register<IDecoratedType, DecoratedType>(Lifetime.Singleton);
             builder.RegisterDecorator<IDecoratedType, Decorator1>();
             builder.RegisterDecorator<IDecoratedType, Decorator2>();
+            
+            var container = builder.Build();
+            var instance = container.Resolve<IDecoratedType>();
+            Assert.That(instance, Is.TypeOf<Decorator2>());
+            
+            var inner2 = ((Decorator2)instance).Inner;
+            Assert.That(inner2, Is.TypeOf<Decorator1>());
+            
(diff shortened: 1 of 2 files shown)


CncealDVRflectN:
Merged changes from `master` and re-targeted the PR to it. Still not sure if all cases are covered.

CncealDVRflectN:
Closing for now as I've found another few cases where it doesn't work. And I think that a different approach with more changes to the codebase is required to implement this feature properly. I will try to write a more detailed explanation in a few days.

LovorDev:
I will wait for your changes :) Let me know if you decide not to finish this.

CncealDVRflectN:
So, the problem I've encountered is that a decorator gets a different instance of a decorated service. Let's say that we register `Service` as a singleton, and it implements `IInitializable` and `IService`. Then we register a decorator for `IService`. That decorator will get a different instance of `Service` than what `IInitializable` resolves to, even though it is registered as a singleton. Also that instance of service, which was injected into the decorator, will not receive a call to `IInitializable.Initialize`.

Here is the test case:

```csharp
[Test]
public void Decorator_Gets_Same_Singleton_Instance_As_Initializable()
{
    var builder = new ContainerBuilder();
    builder.Register<Mocks.InitializableDecoratedType>(Lifetime.Singleton).AsImplementedInterfaces();
    builder.RegisterDecorator<Mocks.IDecoratedType, Mocks.Decorator1>();
    
    using var container = builder.Build();
    var initializableInstance = container.Resolve<IInitializable>();
    var instance = container.Resolve<Mocks.IDecoratedType>();
    Assume.That(instance, Is.TypeOf<Mocks.Decorator1>());
    
    var concreteInstance = (Mocks.Decorator1)instance;
    Assert.That(initializableInstance, Is.EqualTo(concreteInstance.Inner));
}
```

This behavior is caused by the fact that `Registration` is used as a key in several places in the codebase, and it does not implement custom `GetHashCode` and `Equals` methods, so by default any two instances of `Registration` are considered not equal. Current implementation of decorators uses `RegistrationBuilder.Build` method to get the registration for the decorated interface. But every call to this method returns a new instance of `Registration`.

So, here are some approaches to fix this problem:

1. Cache registration on first call to `RegistrationBuilder.Build`. Since you can easily get registration builders from `IContainerBuilder` caching should be done on registration builder level which makes it very error-prone since everyone who implements a new registration builder should know about this problem and implement caching. So, personally, I don't consider this approach as a good one.

2. Implement custom `GetHashCode` and `Equals` methods for `Registration` class. In my opinion, the main problems with this approach are performance, since we need to calculate hash and check for equality the list of interface types, and equality of instance provider. So, this approach also seems like not a good idea.

3. Pass additional context object into `RegistrationBuilder.Build` method which provides some functionality to retrieve existing registrations. Currently, it is my favorite approach, but it also isn't without drawbacks:
   - Breaking registration builder API.
   - Simple implementation will be incompatible with current approach to `VCONTAINER_PARALLEL_CONTAINER_BUILD` inside `ContainerBuilder.BuildRegistry` method since some registration builders can require previous registrations to be built. But if needed, this functionality can also be implemented.
   - In simple implementation some registration builders (decorators included) can trigger linear search of a dependent registration, which can be slow for huge containers. But this also can be solved.

@hadashiA What is your opinion? Maybe you have some other ideas?

PR #800, opened 2025-08-06
BugFix: Not showing the "with key" text in exception if there is no key

migus88:
If there is an error while resolving a registration we would receive a message like this:

> No such registration of type SomeType with Key:

This PR removes the "with Key:" text if there is no key


hadashiA:
👍

Issue #765, opened 2025-04-13
Sometimes, the exception handler fails to register with RegisterEntryPointExceptionHandler()


ganaware:
Here are two new test cases.
The first one simply swaps the registration order within InitializeExceptionHandler().
The second one omits the call to RegisterEntryPoint() within InitializeExceptionHandler().
In both tests, you can confirm that the exception handler was not registered correctly.

```diff
diff --git a/VContainer/Assets/Tests/Unity/EntryPointTest.cs b/VContainer/Assets/Tests/Unity/EntryPointTest.cs
index fd6422b..2693694 100644
--- a/VContainer/Assets/Tests/Unity/EntryPointTest.cs
+++ b/VContainer/Assets/Tests/Unity/EntryPointTest.cs
@@ -71,6 +71,41 @@ namespace VContainer.Tests.Unity
             Assert.That(handled, Is.EqualTo(1));
         }
 
+        [UnityTest]
+        public IEnumerator InitializeExceptionHandlerOrderSwapped()
+        {
+            var handled = 0;
+
+            LifetimeScope.Create(builder =>
+            {
+                builder.RegisterEntryPointExceptionHandler(ex => handled += 1); // Unexpected result: This line was ignored.
+                builder.RegisterEntryPoint<InitializableThrowable>();
+            });
+
+            yield return null;
+            yield return null;
+
+            Assert.That(handled, Is.EqualTo(1)); // Unexpected result: This test did not reach here
+        }
+
+        [UnityTest]
+        public IEnumerator InitializeExceptionHandlerWithoutRegisteringEntryPoint()
+        {
+            var handled = 0;
+
+            var ls = LifetimeScope.Create(builder =>
+            {
+                builder.RegisterEntryPointExceptionHandler(ex => handled += 1); // Unexpected result: This line was ignored.
+            });
+
+            yield return null;
+            yield return null;
+
+            Assert.That(handled, Is.EqualTo(0));
+            ls.Container.Resolve<EntryPointExceptionHandler>().Publish(new System.Exception()); // Unexpected result: The exception was thrown here.
+            Assert.That(handled, Is.EqualTo(1)); // Unexpected result: This test did not reach here
+        }
+
         [UnityTest]
         public IEnumerator PostInitializeExceptionHandler()
         {
```

ganaware:
I confirmed that this issue has been fixed in v1.19.0. Thank you!

PR #804, opened 2025-08-29
Fix problem with RegisterEntryPoint overwriting existing ExceptionHandler

hadashiA:


DenisPimenov:
Hi @hadashiA ! I wanted to ask if there are any reasons why this PR remains open? We have our own logging system where unity logger can connect as a provider. And in environments such as server build, where we need to log to stdout, unhandled exceptions do not get captured by our logging system. Initially, we thought that registering an exception handler was a global setting, but even if we register it in every scope, it still goes to the unity logger. Maybe you could suggest an alternative solution to fix this problem.

DenisPimenov:
I think https://github.com/hadashiA/VContainer/issues/765 has the same reason.

hadashiA:
I'm sorry. I simply haven't been able to spare enough time, so it took me a while to check everything. ><

PR #818, opened 2025-11-15
Add author to package.json

### LINKED ISSUE #731: Missing author description in package json
First of all thank you @hadashiA and all other contributors for for your work! We really benefit from it!

In Unity 6, packages are now shown in the package manager organized by the package author. Since the VContainer package, which we import via scoped registry and openupm, is missing the author field in its package-json, it is sorted under "Packages - Other" (see screenshot).

If the author would be set to Cysharp, it could be listed with other Cysharp packages.

image

Y-YoL:
fix #731

Display the author's name for packages shown on UPM

image


hadashiA:
👍

PR #840, opened 2026-02-17
Prevent VContainer and user code identifier clash

battleroy:
I have encountered the PreserveAttribute types clash (defined in the VContainer namespace and the one that defined inside the namespace of the type, for which the injector was generated). Thus, there I have fixed particularly my case and tried to fix similar possible issues.


hadashiA:
👍

Issue #794, opened 2025-07-22
1.17.0 Unity RegisterComponent with Keyed Throws Exception


oguzhanakpulat:
TestComponent
`builder.RegisterComponent(_testComponent).Keyed(EComponent.Test);`

It throws

`VContainerException: No such registration of type: TestComponent
VContainer.ScopedContainer.Resolve (System.Type type, System.Object key) (at Assets/VContainer/Runtime/Container.cs:93)
VContainer.IObjectResolverExtensions.Resolve[T] (VContainer.IObjectResolver resolver, System.Object key) (at Assets/VContainer/Runtime/IObjectResolverExtensions.cs:10)
VContainer.Unity.ContainerBuilderUnityExtensions+<>c__7`1[TInterface].<RegisterComponent>b__7_0 (VContainer.IObjectResolver container) (at Assets/VContainer/Runtime/Unity/ContainerBuilderUnityExtensions.cs:135)
VContainer.ContainerBuilder.EmitCallbacks (VContainer.IObjectResolver container) (at Assets/VContainer/Runtime/ContainerBuilder.cs:175)
VContainer.ScopedContainerBuilder.BuildScope () (at Assets/VContainer/Runtime/ContainerBuilder.cs:41)
VContainer.ScopedContainer.CreateScope (System.Action`1[T] installation) (at Assets/VContainer/Runtime/Container.cs:126)
VContainer.Container.CreateScope (System.Action`1[T] installation) (at Assets/VContainer/Runtime/Container.cs:263)
VContainer.Unity.LifetimeScope.Build () (at Assets/VContainer/Runtime/Unity/LifetimeScope.cs:199)
VContainer.Unity.LifetimeScope.Awake () (at Assets/VContainer/Runtime/Unity/LifetimeScope.cs:145)`

migus88:
The exception is thrown when resolving a registration. Can you share the code that requests the injection of this component as well as its signature (class name, inheritance...)

oguzhanakpulat:
All components are in same gameobject

```
public class ColorLifetimeScope : LifetimeScope
{
    [SerializeField] private Fsm _fsm;

    protected override void Configure(IContainerBuilder builder)
    {
        builder.RegisterComponent(_fsm).As<IFsm>().Keyed(EFsm.Color);
    }
}
```
```
public enum EFsm
{
    None = 0,
    Color = 1,
    Brush = 2
}
```
```
public class Fsm : MonoBehaviour, IFsm
{
}
```
```
[...]

PR #842, opened 2026-02-25
Fix: Unity RegisterComponent with Keyed Throws Exception

Selakz:
This pull request is related to issue #794. 

**The Problem:**
In the current implementation of `RegisterComponent` and `RegisterComponentInHierarchy`, a build callback is registered to force injection by calling `container.Resolve()`. However, it ignores the key that might be subsequently configured via `.Keyed(key)`. This leads to a `VContainerException` during the build phase because the container tries to resolve a default (unkeyed) registration that doesn't exist. Also, the keyed registration will not be force injected.

Though the main purpose of `RegisterComponent` is to trigger auto-injection, since it returns a `RegistrationBuilder`, I believe ensuring it respects the key parameter is important for internal state consistency.

**The Solution:**
Passed `registrationBuilder.Key` into the `Resolve` method within the build callback. Since the default value of the `key` parameter in `Resolve` is null, I think this change is backward compatible and correctly handles both keyed and non-keyed registrations.


hadashiA:
👍 thx!

PR #845, opened 2026-03-28
fix: resolve GetInstanceID deprecation in Unity 6.4+

chikacc:
## Description

Unity 6.4 has deprecated `Object.GetInstanceID()` (warning), and it will be treated as a compilation error (CS0619) from Unity 6.5 onwards.

This PR addresses the issue by:
- Using `GetEntityId()` for Unity 6.4 and newer versions.
- Maintaining backward compatibility for older Unity versions using conditional compilation (`#if UNITY_6000_4_OR_NEWER`).

**Changes:**

- Modified `LifetimeScope.cs` to use the appropriate unique identifier API based on the Unity version.


hadashiA:
thx!

PR #846, opened 2026-04-03
Add `INonLazy` interface for immediate instantiation via `RegisterEntryPoint`

Ceeeeed:
This PR introduces a new approach to support non-lazy initialization, inspired by previous discussions around the `NonLazy()` method. Instead of using a separate `NonLazy()` API, I provide an `INonLazy` marker interface.  

Perhaps this approach will be more aligned with the repository owner's preferences. For reference, see the previous discussion: [PR #574](https://github.com/hadashiA/VContainer/pull/574).

## Key Points
- Types implementing `INonLazy` will be automatically instantiated when registered via `RegisterEntryPoint`, ensuring immediate construction.
- This avoids the need for empty `Initialize()` implementations solely to trigger early instantiation.
- Follows the existing pattern of `RegisterEntryPoint()`, maintaining API consistency and clarity.
- Includes unit tests demonstrating proper instantiation order and behavior.

## Example Usage

```csharp
public class MyListener : INonLazy
{
    public MyListener()
    {
        EventBus.Subscribe<MyEvent>(OnEvent);
    }

    private void OnEvent(MyEvent e)
    {
        // handle event
    }
}

// Registration
container.RegisterEntryPoint<MyListener>();
// MyListener is created immediately after container build

### DIFF
--- a/VContainer/Assets/Tests/Unity/Fixtures/SampleEntryPoint.cs
+++ b/VContainer/Assets/Tests/Unity/Fixtures/SampleEntryPoint.cs
@@ -165,4 +165,14 @@ namespace VContainer.Tests.Unity
             => throw new System.NotImplementedException();
     }
 #endif
+
+    public class SampleNonLazyEntryPoint : INonLazy
+    {
+        public static int InstanceCount;
+
+        public SampleNonLazyEntryPoint()
+        {
+            InstanceCount++;
+        }
+    }
 }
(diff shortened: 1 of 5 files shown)


hadashiA:
- Question: What does that instance do if it has no Initialize method and is not DIed into anything? If the act of instantiation itself holds meaning, does that mean it's performing some side-effect in the constructor? If so, what is the reason for not making that explicitly an Initialize method?
- If INonLazy has no actual functionality, I think it's better to avoid polluting class definitions as much as possible, similar to other lifetime specifications. Therefore, I prefer the approach in #574.
- Regarding the name "INonLazy":
    - It is not self-evident within this library what "Lazy" refers to. (This term is not used elsewhere.)
    - Rather than using a negation like "Non-", I think giving this feature a proper specific name would be clearer.
        - However, since I am not entirely sure what this feature is intended for, I don't have any good naming ideas at the moment. Perhaps something like "BackgroundService"??

PR #854, opened 2026-05-22
Make method injection parent-first

baijifeilong:


hadashiA:
Rejected for the following reasons:


- It is unclear how changing the order is useful.
- Generally, Injection itself should not have side effects, so it should not depend on the order.
- Implementation issues:
    - Includes changes unrelated to the title.
    - There is no particular guarantee for the default order of GetFields/GetMethods, so Reverse does not guarantee anything.
    - If the reflection implementation is modified, the SourceGenerator implementation must also be updated accordingly.

PR #860, opened 2026-07-03
Optimizing the circular dependency detection

pnarimani:
Problem: The circular dependency checker was running in exponential time, checking every single type on every method call.

Changes:
* Memoizing the result for correct types so we don't recalculate them every time. 
* Store the visiting types in a set so we don't have to iterate the entire stack every time.
* Added tests


pnarimani:
@hadashiA Can you please take a look at this PR? It solves a very problematic issue for large repositories. Thanks!

